PR #10, opened 2022-12-22
feat: add --version option

yaegassy:
Hi, Thanks for a great tool (`ruff`, `ruff-lsp`). 🙇

I thought it would be useful to have a `--version` option in `ruff-lsp`, so I added it. If you do not need this PR, please close it.

**AFTER**:

ruff-lsp-version-option


charliermarsh:
Thanks, this is nice!

PR #31, opened 2022-12-24
Do not show rule explanation when hovering on a string that looks like a noqa comment

harupy:
Do not show rule explanation when hovering on a string that looks like a noqa comment:

https://user-images.githubusercontent.com/17039389/209437516-73fce23c-c1b2-4230-8078-5458e776ad0f.mov


charliermarsh:
Perhaps it's acceptable to show the hover anyway in these cases? Like it still seems like it could be potentially useful, even if it's in a string. What do you think?

harupy:
@charliermarsh Agreed :) Let me close this PR

PR #39, opened 2022-12-25
Update neovim config

igorlfs:
nvim-lspconfig now supports ruff-lsp by default, which allows reducing some overhead in the config

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,8 @@ local on_attach = function(client, bufnr)
 end
 
 -- Configure `ruff-lsp`.
+-- See: https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#ruff_lsp
+-- For the default config, along with instructions on how to customize the settings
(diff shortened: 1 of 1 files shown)


charliermarsh:
Oh, amazing, thanks! Can we add a link to https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#ruff_lsp and include the snippet that shows how to provide settings?

(I’m on phone, can do it when I’m back on a computer and merge, or if you want to update that’s cool too and I can merge from here.)

igorlfs:
> Can we add a link to https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#ruff_lsp 

Done!

> include the snippet that shows how to provide settings?

Nvim-lspconfig's repo already has that, so I decided to only reference it.

charliermarsh:
Sounds good, thanks again!

PR #51, opened 2023-01-02
ruff fromEnvironment (vscode extension) not found under windows

FrancescElies:
**Problem:** vscode with ruff strategy `fromEnvironment` throws the following error under windows
```
        Interpreter executable (./venv/py310/Scripts/ruff) not found; falling back to bundled executable
```

Sadly under windows ruff has the `.exe` extension, thus not found.

**This pr** fixes that

### DIFF
--- a/ruff_lsp/server.py
+++ b/ruff_lsp/server.py
@@ -66,7 +66,7 @@ LSP_SERVER = server.LanguageServer(
     max_workers=MAX_WORKERS,
 )
 
-TOOL_MODULE = "ruff"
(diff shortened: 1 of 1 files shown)


FrancescElies:
platform is more redeable thanks!

charliermarsh:
I'm confused why this didn't fail on CI before. I need to look into that.

FrancescElies:
Maybe CI is using the bundle? This failed when switching to `fromEnvironment `

Issue #61, opened 2023-01-27
Support `textDocument/formatting`


timbedard:
It looks like the only way to have ruff-lsp format a document right now is via a `textDocument/codeAction`. From my experience, most LSP servers allow document formatting via `textDocument/formatting`.

It took me a bit to realize, because the example Neovim config registers a keymap for `vim.lsp.buf.format()`, so I was a bit baffled trying to use that to format a file, as ruff-lsp does nothing when that request is issued right now.

kalekseev:
Not sure what the convention for linter lsps and formatting command but at least something like `RuffFixAll` command would be helpful, this is how eslint do it https://github.com/neovim/nvim-lspconfig/blob/d005193b6476088ff032607baca07b1bbbeb36ee/lua/lspconfig/server_configurations/eslint.lua#L165

kalekseev:
For example if someone setup null-ls with black and ruff-lsp that support formatting then he will need to control which lsp should handle formatting command https://github.com/jose-elias-alvarez/null-ls.nvim/wiki/Formatting-on-save#choosing-a-client-for-formatting

kalekseev:
I think RuffFixAll command should be added to the nvim-lspconfig see my link how eslint do this in nvim-lspconfig.

Jackevansevo:
I'm still seeing this issue. Testing in neovim, using the configuration listed here: https://beta.ruff.rs/docs/editor-integrations/#language-server-protocol-official

Running

```
:lua vim.lsp.buf.format()
```

```
[LSP] Format request failed, no matching language servers
```

I can see https://github.com/charliermarsh/ruff-lsp/commit/9f289ec44bee3c535ed8813603d1202722c1094c got reverted, so unsure whether this is a config issue, or something that hasn't been merged yet.

PR #64, opened 2023-02-11
Handle for LSP textDocument/formatting request

PushUpek:
Fixes: #61

Handle for LSP Server textDocumnet/formatting request

### DIFF
--- a/ruff_lsp/server.py
+++ b/ruff_lsp/server.py
@@ -20,6 +20,7 @@ from lsprotocol.types import (
     TEXT_DOCUMENT_DID_CLOSE,
     TEXT_DOCUMENT_DID_OPEN,
     TEXT_DOCUMENT_DID_SAVE,
+    TEXT_DOCUMENT_FORMATTING,
     TEXT_DOCUMENT_HOVER,
     AnnotatedTextEdit,
     ClientCapabilities,
@@ -34,6 +35,7 @@ from lsprotocol.types import (
     DidCloseTextDocumentParams,
     DidOpenTextDocumentParams,
     DidSaveTextDocumentParams,
+    DocumentFormattingParams,
     Hover,
     HoverParams,
     InitializeParams,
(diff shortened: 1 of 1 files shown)


charliermarsh:
I need to understand how ESLint handles this before merging.

charliermarsh:
Interesting, so after this change, Ruff is selectable as a formatter. I wonder why that _doesn't_ happen with ESLint.

Screen Shot 2023-02-13 at 3 49 18 PM

charliermarsh:
Ah, ok -- so they have a special setting that enables it to register as a formatter:

Screen Shot 2023-02-13 at 3 56 07 PM
Screen Shot 2023-02-13 at 3 56 15 PM

We should do the same thing, I think. Although I guess that should be respected by _clients_ so it's not part of the LSP, but the client reference.

rchl:
I suppose one could argue whether ruff really does format the document or just fixes the issues (including ordering imports). 

Since editors currently provide various functionalities to fix all issues at once (including running code actions on save), I'm not really sure if there is benefit in expose that functionality through formatting request.

charliermarsh:
My goal was just to follow ESLint here, so I was going to disable this by default in VS Code and allow users to toggle it on. I guess that has the downside that other clients need to do the same -- but does ESLint not suffer from the same issue? Maybe they're not meant to act as a generic LSP.

rchl:
But eslint has it disabled by default: https://github.com/microsoft/vscode-eslint/blob/705f5f5d86aab57b16c36e31f229549883d6a37f/package.json#L425-L430

It might be a bit of a legacy option because there was no "code actions on save" back when this option was introduced, I'm pretty sure.

charliermarsh:
Right, and Ruff would too, but maybe I misunderstood what's going on? I'd thought that the ESLint LSP server always registered for the event type (https://github.com/microsoft/vscode-eslint/blob/415526ff31aeb866298514384ad401f2e4d841f5/server/src/eslintServer.ts#L1061), but that the client was responsible for enabling and disabling the behavior -- in which case, that wouldn't have any relationship to the LSP (this repo), but to the clients. I'm guessing that's an inaccurate description though?

rchl:
I'm running this server in ST (using https://github.com/sublimelsp/LSP-eslint/) and I've just checked the server logs.

It doesn't register as a formatter when `format` is disabled.
When `format` is enabled, it decides per file whether formatting should be allowed for that file and dynamically registers formatting capability just for this file if so.

charliermarsh:
Ahh ok, interesting. Perhaps we should roll this back then. I don't _think_ `pygls` enables us to conditionally register for capabilities.

rchl:
I don't know `pygls` but I've looked at its code and it seems to provide the [`register_capability` method](https://github.com/openlawlibrary/pygls/blob/3fc13c4bc6eb7438e5b7e8f662d2274cf5f056c6/pygls/server.py#L429-L432). This would use the newer [dynamic capability registration](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#client_registerCapability).

The simpler option would be to just not announce formatting capabilities in the `initialize` response based on provided `initializationOptions` but that way might indeed not be supported by `pygls`.

PR #66, opened 2023-02-14
Update Sublime Text instructions

LDAP:
An LSP helper plugin for ruff is now available.
https://github.com/sublimelsp/LSP-ruff

The helper automatically manages and updates the `ruff-lsp` binary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,24 +100,7 @@ Ruff also integrates with [`coc.nvim`](https://github.com/neoclide/coc.nvim/wiki
 ### Example: Sublime Text
 
 To use `ruff-lsp` with Sublime Text, install Sublime Text's [LSP](https://github.com/sublimelsp/LSP)
-package, then add something like the following to `LSP.sublime-settings`:
-
-```json
-{
-  "clients": {
-    "ruff-lsp": {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -100,24 +100,7 @@ Ruff also integrates with [`coc.nvim`](https://github.com/neoclide/coc.nvim/wiki
...
-      "initializationOptions": {
-        "settings": {
-          "args": []
-        }
-      }
-    }
-  }
-}
-```
+and [LSP-ruff](https://github.com/sublimelsp/LSP-ruff) package.
```
charliermarsh, line 103: Amazing! Can we include an example of how you would configure it, e.g., to change one of the settings?

LDAP: Should I add that to the LSP-ruff Readme instead? This way people coming from package control (the ST package manager) can profit from that too.

LDAP: I added a section to the LSP-ruff readme.

LDAP: In Sublime Text, the settings are displayed as a split view:
image

charliermarsh: Thanks!

PR #68, opened 2023-02-16
Find ruff in PATH as fallback

ruby0b:
ruff might not always be installed in the python interpreter's scripts directory.
I'm in the process of packaging ruff-lsp for nixpkgs and the fact that ruff-lsp only looks in python's script directory makes it very hard to package in the context of nix since that directory will only ever include the default python binaries.

This PR solves that issue by also looking for a ruff executable in the regular binary PATH.

### DIFF
--- a/ruff_lsp/server.py
+++ b/ruff_lsp/server.py
@@ -8,6 +8,7 @@ import os
 import pathlib
 import platform
 import re
+import shutil
 import sys
 import sysconfig
 from typing import Any, Sequence, cast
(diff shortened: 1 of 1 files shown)


jsr-p:
Nice! 
Alternatively, you can specify the `path` key of the [settings](https://github.com/charliermarsh/ruff-lsp#settings) dictionary to the path of your `ruff` executable :smiley:

ruby0b:
I completely missed #52 which solves my issue and other use cases that this one doesn't cover

PR #70, opened 2023-02-17
Recommend LSP-ruff package for Sublime Text users

jfcherng:
`LSP-ruff` is a helper package which manages (install/upgrade) `ruff-lsp` for the user. It may provide more custom features if needed.


jfcherng:
close as I just notice https://github.com/charliermarsh/ruff-lsp/pull/66

PR #76, opened 2023-02-25
Add "Disable for this line" quickfix

### LINKED ISSUE #50: [feature] quick fix to quickly ignore an error
With ESlint's VSCode's integration there is an ability to quick fix ignore an error:
image

image

Currently with ruff this isn't available 

image

Thank you for ruff, it's great!

bluetech:
Fix #50.

This likely has some unhandled edge cases, but hopefully better than nothing :)

### DIFF
--- a/ruff_lsp/server.py
+++ b/ruff_lsp/server.py
@@ -132,10 +132,13 @@ def _parse_output_using_regex(content: str) -> list[Diagnostic]:
     #   {
     #     "code": "F841",
     #     "message": "Local variable `x` is assigned to but never used",
-    #     "fixed": false,
     #     "location": {
     #       "row": 2,
     #       "column": 5
+    #     },
+    #     "end_location": {
+    #       "row": 2,
+    #       "column": 6
     #     },
     #     "fix": {
     #       "content: "",
@@ -148,7 +151,8 @@ def _parse_output_using_regex(content: str) -> list[Diagnostic]:
     #         "column: 0
     #       }
     #     },
-    #     "filename": "/path/to/test.py"
+    #     "filename": "/path/to/test.py",
+    #     "noqa_row": 2
     #   },
     #   ...
     # ]
@@ -167,7 +171,11 @@ def _parse_output_using_regex(content: str) -> list[Diagnostic]:
             severity=_get_severity(check["code"]),
             code=check["code"],
             source=TOOL_DISPLAY,
-            data=check.get("fix"),
+            data=DiagnosticData(
+                fix=check.get("fix"),
+                # Available since Ruff v0.0.TBD.
+                noqa_row=check.get("noqa_row"),
+            ),
             tags=_get_tags(check["code"]),
         )
         diagnostics.append(diagnostic)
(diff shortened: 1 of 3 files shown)


ruff_lsp/server.py
```diff
@@ -421,6 +423,47 @@ def code_action(params: CodeActionParams) -> list[CodeAction] | None:
                         ),
                     )
 
+    # Add "Disable for this line" for every diagnostic.
```
charliermarsh, line 426: We probably need to return the `noqa_line` for each diagnostic, since this won't work correctly for multi-line strings.

bluetech: Thanks for the pointer, I'll try adding it.

ruff_lsp/server.py
```diff
@@ -416,9 +431,50 @@ def code_action(params: CodeActionParams) -> list[CodeAction] | None:
...
+                            edit=_create_workspace_edit(document, fix),
+                            diagnostics=[diagnostic],
+                        ),
+                    )
+
+    # Add "Disable for this line" for every diagnostic.
+    if not params.context.only or CodeActionKind.QuickFix in params.context.only:
+        for diagnostic in params.context.diagnostics:
+            if diagnostic.source == "Ruff":
+                noqa_row = cast(DiagnosticData, diagnostic.data)["noqa_row"]
```
charliermarsh, line 443: I think this needs to be more defensive, since this key could be absent for older versions of Ruff (as opposed to being `None`).

ruff_lsp/server.py
```diff
@@ -416,9 +431,50 @@ def code_action(params: CodeActionParams) -> list[CodeAction] | None:
...
+                if noqa_row is not None:
+                    line = document.lines[noqa_row - 1].rstrip("\r\n")
+                    match = NOQA_REGEX.search(line)
+                    # `foo  # noqa: OLD` -> `foo  # noqa: OLD,NEW`
+                    if match and match.group("codes") is not None:
+                        codes = match.group("codes") + f",{diagnostic.code}"
```
charliermarsh, line 449: Nit: lets add a space after the comma, to be consistent with Ruff's own `RUF100` behavior.

ruff_lsp/server.py
bluetech: The DiagnosticData is created by the LSP, when I create it I use `noqa_row=check.get("noqa_row")` so I think it's OK?

ruff_lsp/server.py
```diff
@@ -416,9 +431,50 @@ def code_action(params: CodeActionParams) -> list[CodeAction] | None:
...
+                        start, end = match.start("codes"), match.end("codes")
+                        new_line = line[:start] + codes + line[end:]
+                    # `foo  # noqa` -> `foo  # noqa: NEW`
```
charliermarsh, line 452: I think this case might be impossible... since if a line has `foo  # noqa`, the error will by definition be suppressed, right? And so this action wouldn't appear?

bluetech: I managed it using the "useless noqa" lint :)

ruff_lsp/server.py
bluetech: Added a space after the comma

ruff_lsp/server.py
charliermarsh: Ah you're right!

charliermarsh:
Thx!

charliermarsh:
@bluetech - Out of curiosity, what editor do you use with the LSP?

bluetech:
I use vscode

charliermarsh:
Yeah interesting. It's close to Ruff's `--add-noqa` feature.

PR #77, opened 2023-03-01
doc: add extra pointers to the README

markis:
I think it would be helpful to call out that `settings` go under `init_options`.  It was in the docs linked in the comments, but that took me a little too long to figure out.  Maybe this would help others.


charliermarsh:
Thank you, I agree! Appreciate the PR.

PR #83, opened 2023-03-10
Prefer interpreter's Ruff to environment's Ruff

charliermarsh:
See: https://github.com/charliermarsh/ruff-vscode/issues/141.

### DIFF
--- a/ruff_lsp/server.py
+++ b/ruff_lsp/server.py
@@ -794,31 +794,30 @@ def _executable_path(settings: dict[str, Any]) -> str:
             )
 
         path = os.path.join(INTERPRETER_PATHS[settings["interpreter"][0]], TOOL_MODULE)
-        if bundle and not os.path.exists(path):
-            log_to_output(
-                f"External interpreter executable ({path}) not found; "
-                f"falling back to bundled executable: {bundle}"
-            )
-            path = bundle
-        else:
-            log_to_output(f"Using external interpreter executable: {path}")
+    else:
+        path = os.path.join(sysconfig.get_path("scripts"), TOOL_MODULE)
+
+    # First choice: the executable in the current interpreter's scripts directory.
+    if os.path.exists(path):
+        log_to_output(f"Using interpreter executable: {path}")
         return path
 
-    # If the interpreter is same as the interpreter running this process, get the
-    # script path directly.
-    path = os.path.join(sysconfig.get_path("scripts"), TOOL_MODULE)
+    # Second choice: the executable in the global environment.
     environment_path = shutil.which("ruff")
     if environment_path:
         log_to_output(f"Using environment executable: {environment_path}")
-        path = environment_path
-    elif bundle and not os.path.exists(path):
+        return environment_path
+
+    # Third choice: bundled executable.
+    if bundle:
         log_to_output(
             f"Interpreter executable ({path}) not found; "
             f"falling back to bundled executable: {bundle}"
         )
-        path = bundle
-    else:
-        log_to_output(f"Using interpreter executable: {path}")
+        return bundle
+
+    # Last choice: just return the expected path for the current interpreter.
+    log_to_output(f"Unable to find interpreter executable: {path}")
     return path
 
 


yaegassy:
Starting with ruff-lsp `v0.2.1`, ruff installed in a virtual environment is no longer available. The virtual environment is already activated.

I have a question, Is it correct that `importStrategy` should be set to `fromEnvironment` if I want to use ruff in the virtual environment?

I am using Vim/Neovim, and the `importStrategy` setting may be working in VSCode, but maybe the importStrategy setting is not working as intended in non-VSCode environments.

yaegassy:
Hi, @rchl

Does the `fromEnvironment` setting in `importStrategy` work correctly in `sublimelsp/LSP-ruff`? If you can confirm this, I would appreciate it if you could let me know. <https://github.com/charliermarsh/ruff-vscode#settings>

charliermarsh:
Oh weird, I mean this definitely changed, but I wouldn't have expected it to break anything, since in both cases, we're only returning a valid path to Ruff (we just now prefer the activate interpreter's Ruff over a global Ruff). What behavior are you seeing?

yaegassy:
The current behavior is that the ruff command installed with the ruff-lsp dependency is used.

yaegassy:
I debugged it in code and it seems to match this conditional branch. 

https://github.com/charliermarsh/ruff-lsp/blob/4da5c210df3fed7da9c9ccfdabdbf6e5bd636162/ruff_lsp/server.py#L810-L812

charliermarsh:
Mmm, yeah, I can see why that wouldn't work when `ruff-lsp` is installed in some other virtual environment, and why it _would_ work in VS Code. What we had before was a bug though. We need different logic for the different cases... The thing is, VS Code does some work to ensure that the LSP runs within the current user's virtual environment (and bundles Ruff separately, so Ruff and the LSP aren't installed in that virtual environment, but they are run within it). I'm not sure if other clients have similar capabilities -- we kind of get this "for free" from VS Code, and the use of the LSP outside of VS Code thus has some vestiges of that initial implementation.

We might want to have separate logic for VS Code vs. other LSPs. Or we might want to have three settings here: `fromBundled`, `fromEnvironment`, and `fromGlobal` or something to indicate the `which` usage.

yaegassy:
Thanks for the explanation about VS Code and importStrategy.

In the meantime, the problem I have will be addressed with a patch on the language client and related plugins side.

rchl:
> Does the `fromEnvironment` setting in `importStrategy` work correctly in `sublimelsp/LSP-ruff`? If you can confirm this, I would appreciate it if you could let me know. [charliermarsh/ruff-vscode#settings](https://github.com/charliermarsh/ruff-vscode#settings)

We haven't yet updated to the latest `ruff-lsp` in `LSP-ruff` but when I try it locally it seems to work fine. In our case we always load `ruff-lsp`/`ruff` from the virtual environment created specifically for the the `LSP-ruff` package. So in that case it seems to work the same with both the old and the new version of `rust-lsp`.

yaegassy:
@rchl Thank you for the confirmation and explanation in `sublimelsp/LSP-ruff`. 🙇

My problem is that the default for ruff-lsp is to use the ruff command installed with the ruf-lsp dependency, but since `v0.0.18` it has been changed to use the ruff command if it exists in the environment. This is what I sent a pull request for. <https://github.com/charliermarsh/ruff-lsp/pull/52>

A use case is to create a virtual environment as a project, and if the ruff command exists in that virtual environment, the ruff command will be used.

However, since `v0.0.21`, it has reverted back to its original behavior, which is why I commented.

I asked this question because I thought that something else (importStrategy) might need to be adjusted with this change.

rchl:
Sublime Text takes a fairly simplistic approach of always running ruff-lsp and ruff from a separate environment because it's not really possible to activate project's own python environment reliably as in ST all windows share same process and thus system environment. So changing PATH (for example) in one window would affect all other windows and projects.

So the fact that it doesn't have issue with this change might be a factor of its limitations.

PR #86, opened 2023-03-11
Add `--force-exclude` to default arguments

charliermarsh:
Closes #147.


ruff_lsp/server.py
```diff
@@ -770,6 +778,7 @@ def _get_settings_by_document(document: workspace.Document | None) -> dict[str,
 def _executable_path(settings: dict[str, Any]) -> str:
     """Returns the path to the executable."""
     bundle = get_bundle()
+    print("FOO")
```
rchl, line 781: Should be removed.

charliermarsh: Thank you 🤦

PR #101, opened 2023-05-01
Upgrade Ruff version to v0.0.264

patillacode:
Hi there!

This PR intends to just upgrade the `ruff` version.

As you can see I modified the version in `pyproject.toml` under `dependencies`.

I ran both `pip-tools` commands to generate the `requirements.txt` and the `requirements-dev.txt` files:
- `pip-compile --output-file=requirements.txt --resolver=backtracking pyproject.toml`
- `pip-compile --extra=dev --output-file=requirements-dev.txt --resolver=backtracking pyproject.toml`

I used `python@3.8` since that is the oldest `python` version I currently have in my system. It should not be a problem AFAIK.

----

My motivation for this change:

I am a Sublime Text user _(I know, I know...)_ and I have recently reset my whole editor from scratch since after all these years it had too many plugins installed and my config files were a mess.
I completely made the switch to `ruff` (**awesome tool btw, thanks!**) after finding about it. (that was the main motivation to reset my whole editor tbh!)

To properly use `ruff` I installed the related [sublimelsp/LSP-ruff](https://github.com/sublimelsp/LSP-ruff) plugin which depends directly on this project (`ruff-lsp`).

After a lot of head scratching I figured out why my config wasn't working, it was because some of the features I absolutely need (isort related) are only available on the latest version of `ruff`.
I looked into the code and I saw its [direct dependency](https://github.com/sublimelsp/LSP-ruff/blob/5b904ba537f37026c5704e31a9d28fdf0f497247/requirements.in) with `ruff-lsp`

The plugin installs `ruff-lsp@0.0.24` and `ruff@0.0.253` so I patched it with a soft link for the time being to my local install with `ruff@0.0.263`, although this is far from a solution:

Screenshot 2023-05-02 at 01 38 40

My thinking process is that if `ruff-lsp` releases a `0.0.25` version with `ruff` at the latest version, then the plugin would need to pull the `0.0.25` version and the `ruff` version would be the one with the features I am looking for.

Please let me know if what I am saying is valid or if I got it wrong!


charliermarsh:
Thanks for this! We should indeed upgrade the version. Do you mind re-lock with Python 3.7, or enabling maintainer edits so that I can do the same?

charliermarsh:
As an end-user, an alternative solution is that you can point `ruff-lsp` to another `ruff` binary executable via the [`path`](https://github.com/charliermarsh/ruff-lsp#settings) setting.

PR #113, opened 2023-05-16
Switch to pytest

konstin:
Also updates the test fixtures with the latest ruff build

PR #114, opened 2023-05-16
Add hidden-by-default ruff.applyFormat command

konstin:
This adds a very basic format command that depends on https://github.com/charliermarsh/ruff/pull/4453.


konstin:
* main
  * **PR #113** <a href="https://app.graphite.dev/github/pr/charliermarsh/ruff-lsp/113" target="_blank">Graphite</a> 
    * **PR #114** <a href="https://app.graphite.dev/github/pr/charliermarsh/ruff-lsp/114" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/charliermarsh/ruff-lsp/114?utm_source=stack-comment).

MichaReiser:
I'll have a closer look tomorrow. There's one question that came to mind:
Can you add an explanation to the summary explaining your reasoning behind adding a new format comment over using the format capability https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/?

charliermarsh:
Is the idea that we would eventually support the conventional LSP formatting actions, and that this custom action is temporary? Or that both would exist?

ruff_lsp/server.py
```diff
@@ -53,6 +57,11 @@
 from typing_extensions import TypedDict
 
 from ruff_lsp import __version__, utils
+from ruff_lsp.utils import RunResult
+
+if os.environ.get("RUFF_LSP_DEBUG"):
+    log_file = Path(__file__).parent.parent.joinpath("pygls.log")
```
MichaReiser, line 63: Is this the standard logfile name when using python or what does `pygls` stand for?

ruff_lsp/server.py
```diff
@@ -588,46 +595,76 @@ def apply_organize_imports(arguments: tuple[TextDocument]):
...
     )
 
 
-def _formatting_helper(
-    document: workspace.Document, *, only: str | None = None
+if os.environ.get("RUFF_BETA_INTERNAL"):
+
+    @LSP_SERVER.command("ruff.applyFormat")
+    def apply_format(ls: server.LanguageServer, arguments: tuple[TextDocument]):
+        apply_format_impl(ls, arguments)
```
MichaReiser, line 607: A bundler could be useful here so that we could have two different builds. But I don't see how we could achieve the same with python. I guess what we really want are feature flags and your `RUFF_BETA_INTERNAL` is kind of that

ruff_lsp/server.py
```diff
@@ -588,46 +595,76 @@ def apply_organize_imports(arguments: tuple[TextDocument]):
...
+
+
+def apply_format_impl(ls: server.LanguageServer, arguments: tuple[TextDocument]):
```
MichaReiser, line 610: Nit:
```suggestion
def apply_format_impl(language_server: server.LanguageServer, arguments: tuple[TextDocument]):
```

ruff_lsp/server.py
```diff
@@ -588,46 +595,76 @@ def apply_organize_imports(arguments: tuple[TextDocument]):
...
+    edits = _result_to_edits(document, result)
+    ls.apply_edit(
+        _create_workspace_edits(document, edits),
+        "Ruff: Format code",
+    )
+
+
+def _fix_helper(
+    document: workspace.Document,
+    *,
```
MichaReiser, line 626: What's the meaning of the `*`?

ruff_lsp/server.py
```diff
@@ -955,39 +1032,25 @@ def _run_tool_on_document(
     else:
         argv += [document.path]
 
-    log_to_output(f"Running Ruff with: {argv}")
-    result: utils.RunResult = utils.run_path(
+    return run_path(
         argv=argv,
         use_stdin=use_stdin,
         cwd=settings["cwd"],
-        source=document.source.replace("\r\n", "\n"),
```
MichaReiser, line 963: Is it intentional that you removed the `\r\n` replacement? I don't think it is needed for ruff today and could, potentially, result in incorrect offsets.

MichaReiser (approved):
Looks good to me, assuming that the longterm plan is to replace the custom command with the LSP format capabilities. Would you mind expanding the PR summary with what your long-term plans are (e.g. This PR adds a temporary `ruff.applyFormat` command to ease testing and fast iteration. It is gated by the `ENV_...` and my long-term plan is to replace the command with the LSP-s built in format capabilities.

ruff_lsp/server.py
konstin: `pygls` is the lsp library we're using, the code code is just the example from https://pygls.readthedocs.io/en/latest/pages/advanced_usage.html?highlight=log#logging. I'm happy to change this to anything else if you have good ideas for logging, this was a very opportunistic addition because i sometimes need a log file that's independent of VS code

konstin: Good point, renamed it `ruff-lsp.log`

ruff_lsp/server.py
konstin: Python has extras but they unfortunately don't persist to runtime

ruff_lsp/server.py
konstin: Afterwards keyword argument only, e.g. `_fix_helper(document, only="F")` is valid but `_fix_helper(document, "F")` is not. This part i haven't change, no specific opinion given that those are internal APIs and we can change them at will.

https://stackoverflow.com/questions/59661042/what-do-single-star-and-slash-do-as-independent-parameters

ruff_lsp/server.py
konstin: I asked charlie and he said it's likely historical. I removed it because i would like avoid doing any content transformations on the LSP side, those should all be done in ruff, and i'm a bit worried differences between the file in VS Code/on disk and what ruff thinks the file can get really confusing. For the offsets we communicate line and character in the LSP protocol, so i'd hope that they are unaffected

konstin:
Missing steps:
 * Merge https://github.com/charliermarsh/ruff/pull/4453
 * revert CI changes

konstin:
waiting for the next ruff release to unblock CI

Issue #128, opened 2023-06-05
Code action with non-fixable diagnostics returns an error


dhruvmanila:
So, I was testing this (https://github.com/charliermarsh/ruff/issues/4746) out and I've Neovim configured to show a lightbulb symbol if there are any code actions available. This is done by sending a `textDocument/codeAction` request wherever the cursor is present.

***Request:***
```lua
{
  id = 5,
  jsonrpc = "2.0",
  method = "textDocument/codeAction",
  params = {
    context = {
      diagnostics = { {
          code = "PLE0117",
          data = {
            noqa_row = 8
          },
          message = "Nonlocal name `x` found without binding",
          range = {
            ["end"] = {
              character = 18,
              line = 7
            },
            start = {
              character = 17,
              line = 7
            }
          },
          severity = 2,
          source = "Ruff"
        } }
    },
    range = {
      ["end"] = <1>{
        character = 12,
        line = 7
      },
      start = <table 1>
    },
    textDocument = {
      uri = "file:///Users/dhruv/playground/ruff/src/PLE0117.py"
    }
  }
}
```

***Response:***
```lua
{
  error = {
    code = -32602,
    data = "{'traceback': ['  File \"/Users/dhruv/.local/pipx/venvs/ruff-lsp/lib/python3.11/site-packages/pygls/protocol.py\", line 340, in _handle_request\\n    self._execute_request(msg_id, handler, params)\\n', '  File \"/Users/dhruv/.local/pipx/venvs/ruff-lsp/lib/python3.11/site-packages/pygls/protocol.py\", line 264, in _execute_request\\n    self._send_response(msg_id, handler(params))\\n                                ^^^^^^^^^^^^^^^\\n', '  File \"/Users/dhruv/.local/pipx/venvs/ruff-lsp/lib/python3.11/site-packages/ruff_lsp/server.py\", line 477, in code_action\\n    fix = cast(DiagnosticData, diagnostic.data)[\"fix\"]\\n          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^\\n']}",
    message = "KeyError: 'fix'"
  },
  id = 2,
  jsonrpc = "2.0"
}
```

Extracting the traceback:
```
  File "/Users/dhruv/.local/pipx/venvs/ruff-lsp/lib/python3.11/site-packages/pygls/protocol.py", line 340, in _handle_request
    self._execute_request(msg_id, handler, params)
[...]

PR #129, opened 2023-06-06
Use `get` instead of indexing incase key is absent

dhruvmanila:
fixes: #128


dhruvmanila:
Interesting, `mypy` has chosen to not flag `d[...]` if totality is `False` (https://mypy.readthedocs.io/en/stable/typed_dict.html?highlight=TypedDict#totality)

konstin (approved):

Issue #69, opened 2023-02-16
Not all extra arguments to `ruff` throughs the `args` field in the `settings` are valid


jsr-p:
I was trying to debug `ruff-lsp` by specifying the verbose flag `-v` to the underlying ruff executable.
I did this by following the comment of the readme and the guide [here](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#ruff_lsp).
However, the global `TOOL_ARGS` [here](https://github.com/charliermarsh/ruff-lsp/blob/e1925b0035fa7a9e9e0051e0f4b76d51f4b82731/ruff_lsp/server.py#L75) automatically specifies the `--quiet` flag (to only print lint violations) and therefore the `-v` flag in `args` leads to an error, which is not straight forward to locate in the `LspLog` inside Neovim.
Perhaps the source code should raise an `ValueError` if the `-v` flag is specified in `args`, or maybe the readme should be more upfront that not all `args` to `ruff` are allowed (for obvious reasons; although not so obvious that I understood it the first time :smiley:) ?

PR #145, opened 2023-06-16
Automatically ignore unsupported command-line arguments

charliermarsh:
Closes https://github.com/astral-sh/ruff-lsp/issues/69.

Closes https://github.com/astral-sh/ruff-vscode/issues/174.

### DIFF
--- a/ruff_lsp/server.py
+++ b/ruff_lsp/server.py
@@ -89,6 +89,8 @@ LSP_SERVER = server.LanguageServer(
 
 TOOL_MODULE = "ruff.exe" if platform.system() == "Windows" else "ruff"
 TOOL_DISPLAY = "Ruff"
+
+# Arguments provided to every Ruff invocation.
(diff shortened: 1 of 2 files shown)


ruff_lsp/server.py
```diff
@@ -99,6 +101,37 @@
...
+    "--show-settings",
+    "--show-source",
+    "--silent",
+    "--statistics",
+    "--verbose",
+    "-w",
+    "--watch",
+    # Arguments that are not supported at all, and will error when provided.
+    # "--stdin-filename",
+    # "--format",
```
zanieb, line 133: @charliermarsh why are these commented out?

charliermarsh: Because they aren’t flags, they’re positional arguments, so they’re followed by another argument. I figured this wasn’t worth handling since they would error anyway.

PR #152, opened 2023-06-17
Refactor LSP server to use coroutines via `asyncio`

charliermarsh:
## Summary

The pygls documents recommend using [coroutines](https://pygls.readthedocs.io/en/latest/pages/advanced_usage.html#asynchronous-functions-coroutines) to support cancellation. This PR migrates the LSP to use `asyncio`.

## Test Plan

Created a > 50,000 line Python file, and attempted to make repeated edits before and after.

After this change, there is still some lag when editing:

https://github.com/astral-sh/ruff-lsp/assets/1309177/c7e111c8-3cd2-43cb-8b88-665897a38214

But before this change, the lag is _way_ worse, and you often get out-of-order events!

https://github.com/astral-sh/ruff-lsp/assets/1309177/6c08333d-3628-4321-99ff-b856c6da518e

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,6 +46,7 @@ dev = [
   "mypy==1.3.0",
   "pip-tools>=6.13.0,<7.0.0",
   "pytest>=7.3.1,<8.0.0",
+  "pytest-asyncio==0.21.0",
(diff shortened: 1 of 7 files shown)


dhruvmanila (approved):
Nice!

PR #155, opened 2023-06-17
Remove normalizing newline for source document

dhruvmanila:
## Summary

Remove normalizing newline for source document.

## Test Plan

Tested out using CPython codebase with a few files. The diagnostics are being rendered correctly.


charliermarsh (approved):
For posterity: we tried this before, but it broke Windows (reverted in #137). However, now that we're using `asyncio`, and passing bytes in and out of the subprocess, it seems to work as expected.

charliermarsh:
Thanks @dhruvmanila!

PR #186, opened 2023-07-11
Updates to CI workflow

zanieb:
## Changes

- Only run lint checks on the latest Python version in CI to unblock #185 
- Add Python 3.11 to CI
- Bump Python version used for releases to 3.11 (previously 3.7 which is now EOL)

## Test plan

Verify that CI runs correctly on this pull request

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -25,6 +25,7 @@ jobs:
       - name: Install dependencies
         run: just install
       - name: Run checks
+        if: ${{ matrix.python-version == "3.10" }}
         run: just check
       - name: Run tests
         run: just test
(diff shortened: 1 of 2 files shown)


zanieb:
Checked on 3.11: https://github.com/astral-sh/ruff-lsp/actions/runs/5524653011/jobs/10077239553?pr=186
Skipped on 3.10: https://github.com/astral-sh/ruff-lsp/actions/runs/5524653011/jobs/10077239173?pr=186

zanieb:
Well actually to unblock installation of a newer version of Black we'd need to split the linting dependencies into a separate installation category _or_ add a split to the Black version used depending on the Python version which would _not_ require this change.

zanieb:
Actually, pytest dropped support for 3.7 in https://github.com/pytest-dev/pytest/pull/11152 so we cannot run tests in CI even if we separate the linting jobs.

zanieb:
Better handled in https://github.com/astral-sh/ruff-lsp/issues/187

Issue #214, opened 2023-08-11
Source completions not working in Helix


augray:
Using [helix](https://docs.helix-editor.com/title-page.html), autocompletions do not appear to be working.

To reproduce:

- install helix
- configure helix to use ruff-lsp
- execute the following:
- `hx tmp.py`
  + in editor, press `i` to enter edit mode
  + type the following:

```python
some_variable = 1
print(some_var
```

**Expected behavior**:
 I would expect to see a suggestion to complete the variable in the final line to `some_variable`

**Actual behavior**:
I see no suggestion, only an error (`Undefined name 'some_var'`)

**Configuration**:

```bash
$ ruff-lsp --version
ruff-lsp 0.0.35
$ ruff --version
ruff 0.0.284
$ hx --version
helix 23.05 (7f5940be)
$ cat ~/.config/helix/languages.toml
[[language]]
name = "python"
language-server = { command = "ruff-lsp" }
```

zanieb:
Hi!

Ruff does not support source completions at this time — it is intended to be alongside a tool that does. Please see https://github.com/astral-sh/ruff-lsp/issues/209#issuecomment-1668849088 or https://github.com/astral-sh/ruff-lsp/issues/57#issuecomment-1399540768 for more details.

zanieb:
Glad to see you're using Ruff in Helix though! I have been intending to pick up Helix for a while now :)

augray:
Great, thanks for the info! For what it's worth, and for the benefit of anybody else who comes across this issue in the future: I was able to get a workable setup using `pylsp` and some special plugins. Here are some full details:

- install the following with `pip`:

```
python-lsp-ruff==1.5.1
python-lsp-server==1.7.2
```

- in your `pyproject.toml`:

```toml
[tool.pylsp.plugins.ruff]
enabled = true
```

This will cause `pylsp` to use ruff for its linting/static analysis, while `pylsp` handles all the other language server stuff. In general, it seems `pylsp` has a good number of plugins.

aspizu:
For anyone who comes across this, Helix now supports multiple language servers. You can use ruff alongside pyright or pylsp.

Here is my `languages.toml` configuration:

```toml
[[language]]
name = "python"
roots = ["pyproject.toml"]
language-servers = ["pyright", "ruff", "pylsp"]
formatter = { command = "/home/aspizu/.config/helix/python-format.sh", args = [] }
auto-format = true

[language-server.pyright]
[...]

Issue #119, opened 2023-05-22
Use ruff to format the code


tgy:
I've been searching for 1 hour how to use ruff to format the current buffer and haven't found any example of config that works.

[This issue](https://github.com/astral-sh/ruff-lsp/issues/61) has been closed, even though `:lua vim.lsp.buf.format()` does not work as expected.

Could anyone share a config to make ruff format the current buffer? :) Thanks!

pythops:
You may need to use [null-ls](https://github.com/jose-elias-alvarez/null-ls.nvim) plugin.
```
Plug 'jose-elias-alvarez/null-ls.nvim'
```
Then add this configuration
```
lua << EOF
    local null_ls = require("null-ls")
    local augroup = vim.api.nvim_create_augroup("LspFormatting", {})

    null_ls.setup({
        sources = {
            -- formatting
            null_ls.builtins.formatting.ruff,

            -- diagnostics
            null_ls.builtins.diagnostics.ruff,

        },
         on_attach = function(client, bufnr)
                if client.supports_method("textDocument/formatting") then
                    vim.api.nvim_clear_autocmds({ group = augroup, buffer = bufnr })
                    vim.api.nvim_create_autocmd("BufWritePre", {
                        group = augroup,
                        buffer = bufnr,
                        callback = function()
                            vim.lsp.buf.format({ bufnr = bufnr })
                        end,
                    })
                end
            end,
    })
EOF
```

sbinnee:
I have the same issue. It seems that currently only was to use ruff's import sorting is to call code action and select sorting import as described #61. It's too much for simple formatting. I've been relying on venerable isort for the job. It would be nice to see the progress so that I can get rid of isort in the end.

dhruvmanila:
Thanks for opening this issue! Let me give a bit of background into these different LSP capabilities. `textDocument/formatting` request (which is what is sent by `vim.lsp.buf.formatting()` is sent to the server which supports the formatting capabilities and it formats the entire document. On the other hand, `textDocument/codeAction` is used for code fixes to either fix problems (diagnostics) or to refactor code.

[...]

Issue #57, opened 2023-01-22
textDocument/definition is not supported


krims0n32:
I installed ruff-lsp through mason in Neovim but get this message when trying to go to a definition. Any ideas? I configured per the example in the docs.

LspInfo shows ruff-lsp is attached to the buffer and warnings like line too long do show up in the editor.

charliermarsh:
We don't currently support go-to definition, if that's the question underlying the issue :) `ruff-lsp` only handles integration with Ruff, so surfacing errors and providing actions to correct them. It's recommended that you use `ruff-lsp` with some other Python LSP in order to get go-to definition or other behaviors. But please let me know if I'm misinterpreting.

krims0n32:
Gotcha, that was my question :)

PR #215, opened 2023-08-11
Document limitations and intended use-case in README

zanieb:
We have previously explained this in:

- https://github.com/astral-sh/ruff-lsp/issues/214#issuecomment-1675372464
- https://github.com/astral-sh/ruff-lsp/issues/209#issuecomment-1668849088
- https://github.com/astral-sh/ruff-lsp/issues/57#issuecomment-1399540768
- https://github.com/astral-sh/ruff-lsp/issues/119#issuecomment-1595628355

We should cover this in the README to avoid mismatched expectations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,12 @@ Enables Ruff to be used in any editor that supports the LSP, including [Neovim](
 
 For Visual Studio Code, check out the [Ruff VS Code extension](https://github.com/astral-sh/ruff-vscode).
 
+## Using as a Python LSP
+
+ruff-lsp only supports surfacing errors and providing actions to correct them, it does not support behaviors for
+code navigation or completion. It's recommended to use ruff-lsp alongside another Python LSP in order to get
+a complete experience including features like "go-to definition" or auto-complete.
+
 ## Highlights
 
 ### "Quick Fix" actions for auto-fixable violations (like unused imports)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,6 +15,12 @@ Enables Ruff to be used in any editor that supports the LSP, including [Neovim](
 
 For Visual Studio Code, check out the [Ruff VS Code extension](https://github.com/astral-sh/ruff-vscode).
 
+## Using as a Python LSP
+
+ruff-lsp only supports surfacing errors and providing actions to correct them, it does not support behaviors for
+code navigation or completion. It's recommended to use ruff-lsp alongside another Python LSP in order to get
+a complete experience including features like "go-to definition" or auto-complete.
```
charliermarsh, line 22: I'd suggest removing the header so that `Highlights` and `Installation` remove the most prominent section.

What do you think of this, to make it a little more succinct?

> `ruff-lsp` supports surfacing Ruff diagnostics and providing Code Actions to fix them, but is intended to be used alongside another Python LSP in order to support features like navigation and autocompletion.

zanieb: I like that change.

I want to be able to link to a header though.. hm.

charliermarsh: Could we _also_ put this in an FAQ at the bottom that we can link to?

zanieb: I don't really like FAQs and wanted to keep this in the top so it's seen often. Let's just leave out the extra header for now and duplicate it in an FAQ if it gets annoying.

PR #229, opened 2023-09-01
Downgrade typeguard to 3.0.2

charliermarsh:
I'm not sure why dependabot updated this in https://github.com/astral-sh/ruff-lsp/pull/201, but if you try to install our requirements locally, you get a conflict between pygls (which has `^3.0.0`) and typeguard. It passes on CI because we install with `--no-deps`. Huh?

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -74,13 +74,14 @@ tomli==2.0.1
     #   black
     #   build
     #   mypy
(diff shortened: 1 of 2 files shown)


charliermarsh:
@zanieb - Merging to unblock but if you have any idea why dependabot upgraded here I'm all ears :joy:

PR #230, opened 2023-09-01
Use `ruff format` (alpha) command for formatting

charliermarsh:
## Summary

This PR transitions the LSP to use Ruff's formatter via the (alpha) `ruff format` command. It's not yet recommended for production use, but it's more than ready for us to use it internally :)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf
--- a/justfile
+++ b/justfile
@@ -9,12 +9,12 @@ install:
   pip install --no-deps -r requirements-dev.txt
 
 fmt:
-  black ./ruff_lsp ./tests
-  ruff --fix ./ruff_lsp ./tests
+  ruff check --fix ./ruff_lsp ./tests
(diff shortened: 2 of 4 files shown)


justfile
```diff
@@ -9,12 +9,12 @@ install:
...
-  black ./ruff_lsp ./tests
-  ruff --fix ./ruff_lsp ./tests
+  ruff check --fix ./ruff_lsp ./tests
+  ruff format ./ruff_lsp ./tests
 
 check:
-  ruff ./ruff_lsp ./tests
-  black --check ./ruff_lsp ./tests
+  ruff check ./ruff_lsp ./tests
+  ruff format --check ./ruff_lsp ./tests
```
charliermarsh, line 17: The allure of the unified `ruff check`...

MichaReiser (approved):

MichaReiser:
:partying_face:

.gitattributes
```diff
@@ -0,0 +1 @@
+* text=auto eol=lf
```
charliermarsh, line 1: To be revisited :)

PR #242, opened 2023-09-11
Bump version to 0.0.39 with Ruff 0.0.288

zanieb:


MichaReiser (approved):

charliermarsh (approved):
I don't believe it's necessary to bump the Ruff version and publish a new LSP version, since Ruff itself isn't bundled and the Ruff version isn't pinned, so IIUC this only really affects our local development (since, again IIUC, users end up getting the version specified in the `pyproject.toml`, not the locked requirements files). I'll merge and release anyway.

Issue #210, opened 2023-08-08
Improvement Idea: sort code actions by specificity


MrGreenTea:
Hi, thanks for this great integration. It works really well.

When I have an unused import and want to auto remove it I get a list similar to this:

![image](https://github.com/astral-sh/ruff-lsp/assets/9407731/16bdc609-385f-4f05-88b0-c8fd6e298f76)

In my opinion I would prefer the cursor/line specific code action to be at position 1 as it applies directly to the current cursor position.
I believe the order should be
1. Remove unused import _(currently at 3)_
2. Disable for this line _(currently at 4)_
3. Organize Imports _(currently at 1)_
4. Fix All (_(currently at 2)_

I've taken a look at the server implementation and it seems the order comes from the return of [`code_action` in `server.py`](https://github.com/astral-sh/ruff-lsp/blob/5cc6f02fb885b30dd294624471a7f19636ae86a4/ruff_lsp/server.py#L421).

If first the `Ruff: Autofix` and `"Disable for this line"` are added to the list and then `Organize Imports` and `Fix All` it looks like this:

![image](https://github.com/astral-sh/ruff-lsp/assets/9407731/81275f2c-04ab-4c22-9e4c-ffb04560fae4)

What is your opinion on this? Would you be interested in me providing these changes as a PR?

MrGreenTea:
I was only thinking about the order of the autofix and the suggestion to disable it. I think the way it is right now makes more sense, where the autofixes are shown first and then in a separate "block" the disables. But maybe you think different?

zanieb:
I like the idea! Would you be interested in opening a pull request? @charliermarsh may have an opinion as well.

dhruvmanila:
This is an interesting and useful idea but a point to be noted is that the client _could_ alter the order of code actions. The server may provide the code actions in some order but the client implementation (in this case the Neovim LSP client + Telescope fuzzy finder) might alter the order to display the code actions. In an ideal scenario, the client wouldn't update the order but it's a possibility.

MrGreenTea:
> I like the idea! Would you be interested in opening a pull request? @charliermarsh may have an opinion as well.

Yes I will create a PR for this in the next days. Glad you like the idea as well.


[...]

PR #245, opened 2023-09-18
Sort code actions by specificity

MrGreenTea:
## Summary

It changes the sort order sent to clients. This way code actions for specific diagnostic codes are suggested before more general ones like 'Fix: All' or 'Organize Imports'.

See issue #210 for discussions

## Test Plan

❔ I need help with this how you suggest to do it.


charliermarsh:
Thanks! This makes sense.

charliermarsh:
(I think this is small enough that we don't need dedicated tests for it. I'd typically do some manual testing via the VS Code extension.)

PR #252, opened 2023-09-27
Add a dedicated 'Ruff: Format document' action

charliermarsh:
## Summary

We have this for lint fixing and import sorting. It seems consistent to have one for formatting too.

### DIFF
--- a/ruff_lsp/server.py
+++ b/ruff_lsp/server.py
@@ -682,20 +682,29 @@ async def apply_organize_imports(arguments: tuple[TextDocument]):
 
 if RUFF_EXPERIMENTAL_FORMATTER:
 
+    @LSP_SERVER.command("ruff.applyFormat")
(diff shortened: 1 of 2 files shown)


MichaReiser:
Hmm, not sure how I feel about that. I see the consistency but I'm not aware that any other formatter exposes such a command. What's the use case for it?

charliermarsh:
I guess the dedicated "Ruff: Fix all" is nice if you want to run _just_ the Ruff fixes (as opposed to the generic "Fix all"). Perhaps there isn't really a use-case here for the formatter, apart from testing? Being able to run it without configuring it as your default formatter?

MichaReiser:
Yeah, the fix all makes sense (does ruff has it's own `ruff.fixAll` quick fix?

charliermarsh:
Yeah, it defines a custom action in the VS Code client that triggers the server command defined in the LSP.

PR #254, opened 2023-09-28
Require minimum ruff version

konstin:
**Summary** ruff-lsp will crash on start rather than behave weirdly when there is a too low `ruff` version

Fixes https://github.com/astral-sh/ruff/issues/7408

I've bump the minimum to 0.0.291 due to the formatter changes, but also allowed 0.1 already since it should be compatible (CC @zanieb).

When the error occurs in vs-code, we show a little pop-up with a full traceback in the log:

image

There's duplication because we both show the error and don't catch the exception.

**Test Plan** I've tested this by using 0.0.290, which works with the linter by errors with the formatter


ruff_lsp/server.py
```diff
@@ -95,6 +97,7 @@
 
 TOOL_MODULE = "ruff.exe" if sys.platform == "win32" else "ruff"
 TOOL_DISPLAY = "Ruff"
+VERSION_REQUIREMENT = SpecifierSet(">=0.0.291,<0.2.0")
```
konstin, line 100: we're kinda on the intersection of semver and pep440/pep508 but packaging seems convenient

ruff_lsp/server.py
```diff
@@ -1002,7 +1005,22 @@ def _get_settings_by_document(document: workspace.Document | None) -> WorkspaceS
...
+def _find_ruff_binary(settings: WorkspaceSettings) -> str:
+    """Returns the path to the executable, checking that is has the required minimum
+    version."""
+    executable = _find_ruff_binary_path(settings)
+
+    # This call is cached so it should generally be free
+    version = _executable_version(executable)
+    if not VERSION_REQUIREMENT.contains(version, prereleases=True):
+        message = f"ruff {VERSION_REQUIREMENT} required, but {executable} is {version}"
+        log_error(message)
```
charliermarsh, line 1017: Can we try https://pygls.readthedocs.io/en/latest/pages/user-guide.html#show-message?

ruff_lsp/server.py
```diff
@@ -1002,7 +1005,22 @@ def _get_settings_by_document(document: workspace.Document | None) -> WorkspaceS
...
+        raise RuntimeError(message)
```
charliermarsh, line 1018: I think this should be enforced on a per-subcommand basis -- e.g., we should continue to allow older Ruff versions for all the lint commands, but should error if they try to use the formatter.

ruff_lsp/server.py
```diff
@@ -1002,7 +1005,22 @@ def _get_settings_by_document(document: workspace.Document | None) -> WorkspaceS
...
+    log_to_output(f"Found ruff {version} at {executable}")
+    return executable
+
+
+def _find_ruff_binary_path(settings: WorkspaceSettings) -> str:
```
konstin, line 1023: Calling this everytime without caching as it's done atm seems slow

ruff_lsp/server.py
konstin: Seem like `LS_SHOW_NOTIFICATION` is `off` at least in vs code, so i've changed the method to call `LSP_SERVER.show_message` unconditionally because that's an error we need to show in the UI

ruff_lsp/server.py
```diff
@@ -95,6 +97,9 @@
 
 TOOL_MODULE = "ruff.exe" if sys.platform == "win32" else "ruff"
 TOOL_DISPLAY = "Ruff"
+# Don't force users to upgrade ruff if they don't use the linter
+VERSION_REQUIREMENT_LINTER = SpecifierSet(">=0.0.274,<0.2.0")
```
charliermarsh, line 101: Why 0.0.274? I think we can support much older versions.

ruff_lsp/server.py
```diff
@@ -1002,7 +1007,27 @@ def _get_settings_by_document(document: workspace.Document | None) -> WorkspaceS
...
+    executable = _find_ruff_binary_path(settings)
+
+    # This call is cached so it should generally be free
+    version = _executable_version(executable)
+    if not version_requirement.contains(version, prereleases=True):
+        # Put the versions first because the pop-up is small
+        message = (
+            f"ruff {version_requirement} required, but {version} found at {executable}"
+        )
+        show_error(message)
```
charliermarsh, line 1024: Did this end up working?

ruff_lsp/server.py
```diff
@@ -81,7 +82,8 @@
 GLOBAL_SETTINGS: UserSettings = {}
 WORKSPACE_SETTINGS: dict[str, WorkspaceSettings] = {}
 INTERPRETER_PATHS: dict[str, str] = {}
-EXECUTABLE_VERSIONS: dict[str, str] = {}
+# path to last modified and version
+EXECUTABLE_VERSIONS: dict[str, tuple[float, Version]] = {}
```
charliermarsh, line 86: Can we use a `NamedTuple` for the versions?

ruff_lsp/server.py
konstin: yep, see screenshot in the PR description

ruff_lsp/server.py
konstin: that was the version in pyproject.toml

charliermarsh: Let's change this to something much lower. There's no reason not to support older versions, there's nothing stopping us. (We can change it in `pyproject.toml` too if you'd like.)

`--force-exclude` was added in v0.0.189, so lets set that as the minimum IMO.

konstin: fxied

charliermarsh (approved):

ruff_lsp/utils.py
```diff
@@ -57,9 +59,11 @@ def scripts(interpreter: str) -> str:
     )
 
 
-def version(executable: str) -> str:
+def version(executable: str) -> Version:
     """Returns the version of the executable at the given path."""
-    return subprocess.check_output([executable, "--version"]).decode().strip()
+    output = subprocess.check_output([executable, "--version"]).decode().strip()
+    version = output.replace("ruff ", "")  # no removeprefix in 3.7 :/
```
charliermarsh, line 65: Maybe instead:

```python
if output.startswith("ruff "):
  output = output[len("ruff "):]
```

charliermarsh:
Gonna merge this, I want to work on something that will depend on it. Hope that's cool.

ruff_lsp/server.py
```diff
@@ -1095,15 +1139,17 @@ async def _run_check_on_document(
     # If we're trying to run a single rule, add it to the command line, and disable
     # all other rules (if the Ruff version is sufficiently recent).
     if only:
-        if _executable_version(executable) >= "0.0.198":
+        if VERSION_REQUIREMENT_ALL_SELECTOR.contains(
```
charliermarsh, line 1142: @konstin -- I re-added this, since we now allow linter versions earlier than 0.0.198.

PR #255, opened 2023-10-02
Move args and run settings to `lint.args` and `lint.run`

charliermarsh:
## Summary

This PR moves the `args` and `run` settings under a `lint` namespace, in a backwards-compatible way, such that we continue to respect existing configurations.

These settings only affect the linter and not the formatter. In particular, this will enable us to add `format.args` in the future, for `ruff format` command-line arguments.


ruff_lsp/settings.py
```diff
@@ -2,16 +2,15 @@
 
 from typing_extensions import Literal, TypedDict
 
+Run = Literal["onSave", "onType"]
```
konstin, line 5: Maybe `RunOn` or `RunTrigger` or something else less generic

konstin (approved):

PR #263, opened 2023-10-05
Rename deprecated function / types

dhruvmanila:
## Summary

This PR renames the deprecated symbols for `pygls` in `v1.1.0`.

PR #264, opened 2023-10-05
Add support for Jupyter Notebook

dhruvmanila:
## Summary

This PR adds support for Jupyter Notebook. It requires client support for LSP 3.17 which contains the [Notebook support](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#notebookDocument_synchronization).

#### Context

* `Document`: LSP type representing a text file (Python file for Ruff).
* `TextDocument`: `pygls` representation of the LSP `Document`. This is an abstraction created from a `Document` which provides some useful methods like getting the file path, source code, etc.
* New in 3.17: `NotebookDocument` type was added to represent a Notebook which consists of a list of cells (`NotebookCell`). Note that these are all LSP types coming from `lsprotocol`.
* In `pygls`, a Notebook cell is represented as a text document (`TextDocument`).

There are methods provided by `pygls` to get the object:
* `get_text_document` - Returns a `TextDocument` which either represents a Python file or a Notebook cell
* `get_notebook_document` - Returns a `NotebookDocument` either using the Notebook URI or a cell URI. For cell URI, it returns the `NotebookDocument` containing the cell.

#### Document

A new `Document` type was created to facilitate the implementation. This represents either a Python file, a Notebook or a Notebook cell. There are various constructor methods which should be used to create this type:
* For a URI representing a Python file, use either `from_uri` or `from_text_document`.
* For a URI representing a Notebook file, use either `from_uri` or `from_notebook_document`.
* For a URI representing a Notebook cell, use either `from_cell_or_text_uri` or `from_notebook_cell`.

#### Notebook JSON

Ruff expects the source content of a Notebook file to be in JSON format following the [Notebook format specification] but the protocol uses it's own abstraction and doesn't store the JSON format. This means that we need to create a JSON string representing the Notebook from the available information. This doesn't need all the information as Ruff only uses the cell source and version information. So, we create a minimal JSON string representing the Notebook document and pass it to Ruff.

<details><summary>An example JSON string representing a Notebook Document:</summary>
<p>

```json
{
  "metadata": {},
  "nbformat": 4,
  "nbformat_minor": 5,
  "cells": [
    {
      "cell_type": "code",
      "metadata": null,
      "outputs": [],
      "source": "import random\nimport math"
    },
    {
      "cell_type": "code",
      "metadata": null,
      "outputs": [],
      "source": "try:\n    print()\nexcept ValueError:\n    pass"
    },
    {
      "cell_type": "code",
      "metadata": null,
      "outputs": [],
      "source": "import random\nimport pprint\n\nrandom.randint(10, 20)"
    },
    {
      "cell_type": "code",
      "metadata": null,
      "outputs": [],
      "source": "foo = 1\nif foo is 1:\n    msg = f\"Invalid foo: {foo}\"\n    raise ValueError(msg)"
    }
  ]
}
```

</p>
</details> 

**We need to pass in every cell including the markdown cell to get an accurate information like the cell number.**

For the cell document kind, the source value is a JSON string containing just a single code cell. This is required as code actions and formatting work at both cell and notebook level.

### Configuration

For VSCode users, the `notebook.*` configuration is used to run the formatter or code actions on save:

```jsonc
{
  // Enable formatting the entire Notebook on save
  "notebook.formatOnSave.enabled": true,
  // Run the enabled code actions on the entire Notebook on save
  "notebook.codeActionsOnSave": {
    "source.fixAll": true,
    "source.organizeImports.ruff": true
  },
}
```

The way the above settings work in VSCode is that the editor runs the actions in parallel for every cell. This has the illusion that it was run on the entire Notebook. The commands defined by us (`Ruff: Organize imports` and `Ruff: Fix all auto-fixable problems`) are run on the entire Notebook at once. This is important because in the latter case the `ruff` command is invoked `n` number of times where `n` is the number of cells while for the former it's run only once.

#### Builtin

* `Ruff: Organize Imports`: Works at Notebook level
* `Ruff: Fix all auto-fixable problems`: Works at Notebook level

#### VSCode specifc

* `Format Cell`: Formats the current cell
* `Notebook: Format Notebook`: Formats the entire Notebook by running the formatter for every cell
* `Organize Imports`: Runs the `source.organizeImports` code action on every cell in parallel
* `Fix All`: Runs the `source.fixAll` code action on every cell in parallel

## Feature checklist

- [x] Code actions
  - [x] Organize imports
  - [x] Fix all
  - [x] Each fixable diagnostics
  - [x] Disable rule comment
- [x] Code action resolve
- [x] Commands
  - [x] `ruff.applyAutofix`
  - [x] `ruff.applyOrganizeImports`
  - [x] `ruff.applyFormat`
- [x] Diagnostics
  - [x] On open
  - [x] On close
  - [x] On save
  - [x] On change
- [x] Formatting
- [x] Hover

## Test Plan

Manually testing for all the features mentioned above.

### How to run this locally?

1. Clone https://github.com/astral-sh/ruff-lsp and https://github.com/astral-sh/ruff-vscode in the same directory
2. Checkout this branch `git checkout dhruv/notebook` in the `ruff-lsp` repository
3. Install the requirements for both repositories
4. For `ruff-vscode`, uninstall `ruff-lsp` (`pip uninstall --yes ruff-lsp`) as we'd want to use the local version. To install the local `ruff-lsp` version in `ruff-vscode`, follow [Modifying the LSP](https://github.com/astral-sh/ruff-vscode#modifying-the-lsp).
5. Open VSCode from `ruff-vscode` directory -> "Run and Debug" section from the sidebar -> "Debug Extension and Python" config.

This will then open a VSCode development session which can be used to test out the notebook features.

**Test notebooks:**
* Formatting: https://gist.github.com/dhruvmanila/7803e5a3b98c414505384db415a635a0
* Diagnostics, Code actions, Commands: https://gist.github.com/dhruvmanila/54c65870f167a56558d4701f57f53042

**Requires: https://github.com/astral-sh/ruff/pull/7664 which was released in `v0.1.0`**

fixes: #267 
closes: https://github.com/astral-sh/ruff-vscode/issues/256
closes: https://github.com/astral-sh/ruff-vscode/issues/314
closes: https://github.com/astral-sh/ruff-vscode/issues/51

### DIFF
--- a/ruff_lsp/server.py
+++ b/ruff_lsp/server.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import enum
 import json
 import logging
 import os
@@ -10,13 +11,19 @@ import re
 import shutil
 import sys
 import sysconfig
+from collections.abc import Iterable, Mapping
+from dataclasses import dataclass
 from pathlib import Path
-from typing import NamedTuple, Sequence, cast
+from typing import Any, List, NamedTuple, Sequence, Union, cast
 
 from lsprotocol import validators
 from lsprotocol.types import (
     CODE_ACTION_RESOLVE,
     INITIALIZE,
+    NOTEBOOK_DOCUMENT_DID_CHANGE,
+    NOTEBOOK_DOCUMENT_DID_CLOSE,
+    NOTEBOOK_DOCUMENT_DID_OPEN,
+    NOTEBOOK_DOCUMENT_DID_SAVE,
     TEXT_DOCUMENT_CODE_ACTION,
     TEXT_DOCUMENT_DID_CHANGE,
     TEXT_DOCUMENT_DID_CLOSE,
@@ -34,9 +41,13 @@ from lsprotocol.types import (
     Diagnostic,
     DiagnosticSeverity,
     DiagnosticTag,
+    DidChangeNotebookDocumentParams,
     DidChangeTextDocumentParams,
+    DidCloseNotebookDocumentParams,
     DidCloseTextDocumentParams,
+    DidOpenNotebookDocumentParams,
     DidOpenTextDocumentParams,
+    DidSaveNotebookDocumentParams,
     DidSaveTextDocumentParams,
     DocumentFormattingParams,
     Hover,
@@ -45,6 +56,11 @@ from lsprotocol.types import (
     MarkupContent,
     MarkupKind,
     MessageType,
+    NotebookCellKind,
+    NotebookDocument,
+    NotebookDocumentSyncOptions,
+    NotebookDocumentSyncOptionsNotebookSelectorType2,
+    NotebookDocumentSyncOptionsNotebookSelectorType2CellsType,
     OptionalVersionedTextDocumentIdentifier,
     Position,
     Range,
@@ -54,10 +70,11 @@ from lsprotocol.types import (
 )
 from packaging.specifiers import SpecifierSet, Version
 from pygls import server, uris, workspace
-from typing_extensions import TypedDict
+from typing_extensions import Literal, Self, TypedDict, assert_never
 
 from ruff_lsp import __version__, utils
 from ruff_lsp.settings import (
+    Run,
     UserSettings,
     WorkspaceSettings,
     lint_args,
@@ -102,6 +119,18 @@ LSP_SERVER = server.LanguageServer(
     name="Ruff",
     version=__version__,
     max_workers=MAX_WORKERS,
+    notebook_document_sync=NotebookDocumentSyncOptions(
+        notebook_selector=[
+            NotebookDocumentSyncOptionsNotebookSelectorType2(
+                cells=[
+                    NotebookDocumentSyncOptionsNotebookSelectorType2CellsType(
+                        language="python"
+                    )
+                ]
+            )
+        ],
+        save=True,
+    ),
 )
 
 TOOL_MODULE = "ruff.exe" if sys.platform == "win32" else "ruff"
@@ -176,6 +205,149 @@ UNSUPPORTED_FORMAT_ARGS = [
 
 
 ###
+# Document
+###
+
+
+@enum.unique
+class DocumentKind(enum.Enum):
+    """The kind of document."""
+
+    Text = enum.auto()
+    """A Python file or a cell in a Notebook Document."""
+
+    Notebook = enum.auto()
+    """A Notebook Document."""
+
+
+@dataclass(frozen=True)
+class Document:
+    """A document representing either a Python file, a Notebook cell, or a Notebook."""
+
+    uri: str
+    path: str
+    source: str
+    kind: DocumentKind
+    version: int | None
+
+    @classmethod
+    def from_text_document(cls, text_document: workspace.TextDocument) -> Self:
+        """Create a `Document` from the given Text Document."""
+        return cls(
+            uri=text_document.uri,
+            path=text_document.path,
+            kind=DocumentKind.Text,
+            source=text_document.source,
+            version=text_document.version,
+        )
+
+    @classmethod
+    def from_notebook_document(cls, notebook_document: NotebookDocument) -> Self:
+        """Create a `Document` from the given Notebook Document."""
+        path = uris.to_fs_path(notebook_document.uri)
+        if path is None:
+            # `pygls` raises a `Exception` as well in `workspace.TextDocument`.
+            raise ValueError(
+                f"Unable to convert URI to file path: {notebook_document.uri}"
+            )
+        return cls(
+            uri=notebook_document.uri,
+            path=path,
+            kind=DocumentKind.Notebook,
+            source=_create_notebook_json(notebook_document),
+            version=notebook_document.version,
+        )
+
+    @classmethod
+    def from_uri(cls, uri: str) -> Self:
+        """Create a `Document` from the given URI.
+
+        The URI can be a file URI, a notebook URI, or a cell URI. The function will
+        try to get the notebook document first, and if that fails, it will fallback
+        to the text document.
+        """
+        notebook_document = LSP_SERVER.workspace.get_notebook_document(cell_uri=uri)
+        if notebook_document is None:
+            notebook_document = LSP_SERVER.workspace.get_notebook_document(
+                notebook_uri=uri
+            )
+        if notebook_document is None:
+            text_document = LSP_SERVER.workspace.get_text_document(uri)
+            return cls.from_text_document(text_document)
+        else:
+            return cls.from_notebook_document(notebook_document)
+
+    def is_stdlib_file(self) -> bool:
+        """Return True if the document belongs to standard library."""
+        return utils.is_stdlib_file(self.path)
+
+    def is_notebook_file(self) -> bool:
+        """Return True if the document belongs to a Notebook or a cell in a Notebook."""
+        return self.kind is DocumentKind.Notebook or self.path.endswith(".ipynb")
+
+
+SourceValue = Union[str, List[str]]
+
+
+class CodeCell(TypedDict):
+    """A code cell in a Jupyter notebook."""
+
+    cell_type: Literal["code"]
+    metadata: Any
+    outputs: list[Any]
+    source: SourceValue
+
+
+class MarkdownCell(TypedDict):
+    """A markdown cell in a Jupyter notebook."""
+
+    cell_type: Literal["markdown"]
+    metadata: Any
+    source: SourceValue
+
+
+class Notebook(TypedDict):
+    """The JSON representation of a Notebook Document."""
+
+    metadata: Any
+    nbformat: int
+    nbformat_minor: int
+    cells: list[CodeCell | MarkdownCell]
+
+
+def _create_notebook_json(notebook_document: NotebookDocument) -> str:
+    """Create a JSON representation of the given Notebook Document."""
+    cells: list[CodeCell | MarkdownCell] = []
+    for notebook_cell in notebook_document.cells:
+        cell_document = LSP_SERVER.workspace.get_text_document(notebook_cell.document)
+        if notebook_cell.kind is NotebookCellKind.Code:
+            cells.append(
+                {
+                    "cell_type": "code",
+                    "metadata": None,
+                    "outputs": [],
+                    "source": cell_document.source,
+                }
+            )
+        else:
+            cells.append(
+                {
+                    "cell_type": "markdown",
+                    "metadata": None,
+                    "source": cell_document.source,
+                }
+            )
+    return json.dumps(
+        {
+            "metadata": {},
+            "nbformat": 4,
+            "nbformat_minor": 5,
+            "cells": cells,
+        }
+    )
+
+
+###
 # Linting.
 ###
 
@@ -183,38 +355,137 @@ UNSUPPORTED_FORMAT_ARGS = [
 @LSP_SERVER.feature(TEXT_DOCUMENT_DID_OPEN)
 async def did_open(params: DidOpenTextDocumentParams) -> None:
     """LSP handler for textDocument/didOpen request."""
-    document = LSP_SERVER.workspace.get_text_document(params.text_document.uri)
-    diagnostics: list[Diagnostic] = await _lint_document_impl(document)
+    document = Document.from_text_document(
+        LSP_SERVER.workspace.get_text_document(params.text_document.uri)
+    )
+    diagnostics = await _lint_document_impl(document)
     LSP_SERVER.publish_diagnostics(document.uri, diagnostics)
 
 
 @LSP_SERVER.feature(TEXT_DOCUMENT_DID_CLOSE)
 def did_close(params: DidCloseTextDocumentParams) -> None:
     """LSP handler for textDocument/didClose request."""
-    document = LSP_SERVER.workspace.get_text_document(params.text_document.uri)
+    text_document = LSP_SERVER.workspace.get_text_document(params.text_document.uri)
     # Publishing empty diagnostics to clear the entries for this file.
-    LSP_SERVER.publish_diagnostics(document.uri, [])
+    LSP_SERVER.publish_diagnostics(text_document.uri, [])
 
 
 @LSP_SERVER.feature(TEXT_DOCUMENT_DID_SAVE)
 async def did_save(params: DidSaveTextDocumentParams) -> None:
     """LSP handler for textDocument/didSave request."""
-    document = LSP_SERVER.workspace.get_text_document(params.text_document.uri)
-    if lint_run(_get_settings_by_document(document.path)) in ("onType", "onSave"):
-        diagnostics: list[Diagnostic] = await _lint_document_impl(document)
+    text_document = LSP_SERVER.workspace.get_text_document(params.text_document.uri)
+    if lint_run(_get_settings_by_document(text_document.path)) in (
+        Run.OnType,
+        Run.OnSave,
+    ):
+        document = Document.from_text_document(text_document)
+        diagnostics = await _lint_document_impl(document)
         LSP_SERVER.publish_diagnostics(document.uri, diagnostics)
 
 
 @LSP_SERVER.feature(TEXT_DOCUMENT_DID_CHANGE)
 async def did_change(params: DidChangeTextDocumentParams) -> None:
     """LSP handler for textDocument/didChange request."""
-    document = LSP_SERVER.workspace.get_text_document(params.text_document.uri)
-    if lint_run(_get_settings_by_document(document.path)) == "onType":
-        diagnostics: list[Diagnostic] = await _lint_document_impl(document)
+    text_document = LSP_SERVER.workspace.get_text_document(params.text_document.uri)
+    if lint_run(_get_settings_by_document(text_document.path)) == Run.OnType:
+        document = Document.from_text_document(text_document)
+        diagnostics = await _lint_document_impl(document)
         LSP_SERVER.publish_diagnostics(document.uri, diagnostics)
 
 
-async def _lint_document_impl(document: workspace.TextDocument) -> list[Diagnostic]:
+@LSP_SERVER.feature(NOTEBOOK_DOCUMENT_DID_OPEN)
+async def did_open_notebook(params: DidOpenNotebookDocumentParams) -> None:
+    """LSP handler for notebookDocument/didOpen request."""
+    notebook_document = LSP_SERVER.workspace.get_notebook_document(
+        notebook_uri=params.notebook_document.uri
+    )
+    if notebook_document is None:
+        log_warning(f"No notebook document found for {params.notebook_document.uri!r}")
+        return
+
+    document = Document.from_notebook_document(notebook_document)
+    diagnostics = await _lint_document_impl(document)
+
+    # Publish diagnostics for each cell.
+    for cell_idx, diagnostics in _group_diagnostics_by_cell(diagnostics).items():
+        LSP_SERVER.publish_diagnostics(
+            # The cell indices are 1-based in Ruff.
+            params.notebook_document.cells[cell_idx - 1].document,
+            diagnostics,
+        )
+
+
+@LSP_SERVER.feature(NOTEBOOK_DOCUMENT_DID_CLOSE)
+def did_close_notebook(params: DidCloseNotebookDocumentParams) -> None:
+    """LSP handler for notebookDocument/didClose request."""
+    # Publishing empty diagnostics to clear the entries for all the cells in this
+    # Notebook Document.
+    for cell_text_document in params.cell_text_documents:
+        LSP_SERVER.publish_diagnostics(cell_text_document.uri, [])
+
+
+@LSP_SERVER.feature(NOTEBOOK_DOCUMENT_DID_SAVE)
+async def did_save_notebook(params: DidSaveNotebookDocumentParams) -> None:
+    """LSP handler for notebookDocument/didSave request."""
+    await _did_change_or_save_notebook(
+        params.notebook_document.uri, run_types=[Run.OnSave, Run.OnType]
+    )
+
+
+@LSP_SERVER.feature(NOTEBOOK_DOCUMENT_DID_CHANGE)
+async def did_change_notebook(params: DidChangeNotebookDocumentParams) -> None:
+    """LSP handler for notebookDocument/didChange request."""
+    await _did_change_or_save_notebook(
+        params.notebook_document.uri, run_types=[Run.OnType]
+    )
+
+
+def _group_diagnostics_by_cell(
+    diagnostics: Iterable[Diagnostic],
+) -> Mapping[int, list[Diagnostic]]:
+    """Group diagnostics by cell index.
+
+    The function will return a mapping from cell number to a list of diagnostics for
+    that cell. The mapping will be empty if the diagnostic doesn't contain the cell
+    information.
+    """
+    cell_diagnostics: dict[int, list[Diagnostic]] = {}
+    for diagnostic in diagnostics:
+        cell = cast(DiagnosticData, diagnostic.data).get("cell")
+        if cell is not None:
+            cell_diagnostics.setdefault(cell, []).append(diagnostic)
+    return cell_diagnostics
+
+
+async def _did_change_or_save_notebook(
+    notebook_uri: str, *, run_types: Sequence[Run]
+) -> None:
+    """Handle notebookDocument/didChange and notebookDocument/didSave requests."""
+    notebook_document = LSP_SERVER.workspace.get_notebook_document(
+        notebook_uri=notebook_uri
+    )
+    if notebook_document is None:
+        log_warning(f"No notebook document found for {notebook_uri!r}")
+        return
+
+    document = Document.from_notebook_document(notebook_document)
+    if lint_run(_get_settings_by_document(document.path)) not in run_types:
+        return
+
+    cell_diagnostics = _group_diagnostics_by_cell(await _lint_document_impl(document))
+
+    # Publish diagnostics for every code cell, replacing the previous diagnostics.
+    for cell_idx, cell in enumerate(notebook_document.cells):
+        if cell.kind is not NotebookCellKind.Code:
+            continue
+        LSP_SERVER.publish_diagnostics(
+            cell.document,
+            # The cell indices are 1-based in Ruff.
+            cell_diagnostics.get(cell_idx + 1, []),
+        )
+
+
+async def _lint_document_impl(document: Document) -> list[Diagnostic]:
     result = await _run_check_on_document(document)
     if result is None:
         return []
@@ -260,6 +531,7 @@ def _parse_output(content: bytes) -> list[Diagnostic]:
     # Ruff's output looks like:
     # [
     #   {
+    #     "cell": null,
     #     "code": "F841",
     #     "message": "Local variable `x` is assigned to but never used",
     #     "location": {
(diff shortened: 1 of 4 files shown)


dhruvmanila:
* main
  * **PR #263** <a href="https://app.graphite.dev/github/pr/astral-sh/ruff-lsp/263" target="_blank">Graphite</a> 
    * **PR #264** <a href="https://app.graphite.dev/github/pr/astral-sh/ruff-lsp/264" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/astral-sh/ruff-lsp/264?utm_source=stack-comment).

ruff_lsp/server.py
```diff
@@ -175,6 +197,145 @@ class VersionModified(NamedTuple):
...
+        return cls(
+            uri=notebook_document.uri,
+            path=path,
+            kind=DocumentKind.Notebook,
+            source=_create_notebook_json(notebook_document),
+            version=notebook_document.version,
+        )
+
+    @classmethod
+    def from_uri(cls, uri: str) -> Self:
```
karthiknadig, line 255: You should be able to identify cells as they start with `vscode-notebook-cell` and notebooks start with `vscode-notebook` and virtual files with `vscode-vfs`

ruff_lsp/server.py
```diff
@@ -455,11 +726,11 @@ class LegacyFix(TypedDict):
 )
 async def code_action(params: CodeActionParams) -> list[CodeAction] | None:
     """LSP handler for textDocument/codeAction request."""
-    document = LSP_SERVER.workspace.get_text_document(params.text_document.uri)
+    document = Document.from_uri(params.text_document.uri)
 
     settings = _get_settings_by_document(document.path)
 
-    if utils.is_stdlib_file(document.path):
+    if document.is_stdlib_file():
```
karthiknadig, line 733: You can also provide notebook specific code-actions. These are prefixed by namespace `notebook.` in code action kind.

ruff_lsp/server.py
dhruvmanila: Thanks for the suggestion! Correct me if I'm wrong, but I think that is specific to VSCode while `ruff-lsp` is a generic implementation for any editors implementing the LSP protocol.

ruff_lsp/server.py
```diff
@@ -105,6 +122,18 @@ class VersionModified(NamedTuple):
     name="Ruff",
     version=__version__,
     max_workers=MAX_WORKERS,
+    notebook_document_sync=NotebookDocumentSyncOptions(
```
dhruvmanila, line 125: This is required to register support for Notebook in the server.

ruff_lsp/server.py
```diff
@@ -105,6 +122,18 @@ class VersionModified(NamedTuple):
...
+        notebook_selector=[
+            NotebookDocumentSyncOptionsNotebookSelectorType2(
+                cells=[
+                    NotebookDocumentSyncOptionsNotebookSelectorType2CellsType(
+                        language="python"
+                    )
+                ]
+            )
+        ],
+        save=True,
```
dhruvmanila, line 135: We should receive notifications when a Notebook is saved.

ruff_lsp/server.py
```diff
@@ -178,6 +207,149 @@ class VersionModified(NamedTuple):
...
+
+        The URI can be a file URI, a notebook URI, or a cell URI. The function will
+        try to get the notebook document first, and if that fails, it will fallback
+        to the text document.
+        """
+        notebook_document = LSP_SERVER.workspace.get_notebook_document(cell_uri=uri)
+        if notebook_document is None:
+            notebook_document = LSP_SERVER.workspace.get_notebook_document(
+                notebook_uri=uri
+            )
```
dhruvmanila, line 276: We can't use:

```python
LSP_SERVER.workspace.get_notebook_document(notebook_uri=uri, cell_uri=uri)
```

Because the method uses _either_ `notebook_uri` or `cell_uri` where `notebook_uri` is preferred over `cell_uri`. While our implementation will first try the `cell_uri` and then the `notebook_uri`.

ruff_lsp/server.py
```diff
@@ -178,6 +207,149 @@ class VersionModified(NamedTuple):
...
+                    "cell_type": "markdown",
+                    "metadata": None,
+                    "source": cell_document.source,
+                }
+            )
+    return json.dumps(
+        {
+            "metadata": {},
+            "nbformat": 4,
+            "nbformat_minor": 5,
```
dhruvmanila, line 347: The version doesn't matter in this context as Ruff mainly uses it to add the cell ID to maintain consistency when Ruff writes the notebook content back on disk. For VSCode, we aren't going to touch the actual file content.

ruff_lsp/server.py
```diff
@@ -186,38 +358,137 @@ class VersionModified(NamedTuple):
...
-async def _lint_document_impl(document: workspace.TextDocument) -> list[Diagnostic]:
+@LSP_SERVER.feature(NOTEBOOK_DOCUMENT_DID_OPEN)
+async def did_open_notebook(params: DidOpenNotebookDocumentParams) -> None:
+    """LSP handler for notebookDocument/didOpen request."""
+    notebook_document = LSP_SERVER.workspace.get_notebook_document(
+        notebook_uri=params.notebook_document.uri
+    )
+    if notebook_document is None:
+        log_warning(f"No notebook document found for {params.notebook_document.uri!r}")
+        return
```
dhruvmanila, line 407: This shouldn't happen but just in case.

ruff_lsp/server.py
```diff
@@ -186,38 +358,137 @@ class VersionModified(NamedTuple):
...
+    document = Document.from_notebook_document(notebook_document)
+    diagnostics = await _lint_document_impl(document)
+
+    # Publish diagnostics for each cell.
+    for cell_idx, diagnostics in _group_diagnostics_by_cell(diagnostics).items():
+        LSP_SERVER.publish_diagnostics(
+            # The cell indices are 1-based in Ruff.
+            params.notebook_document.cells[cell_idx - 1].document,
+            diagnostics,
+        )
```
dhruvmanila, line 418: The publish diagnostics capability works on a text document which, for a Notebook, is a single cell. We'll group the diagnostics by cell to make a single request for each cell.

ruff_lsp/server.py
```diff
@@ -186,38 +358,137 @@ class VersionModified(NamedTuple):
...
+    document = Document.from_notebook_document(notebook_document)
+    if lint_run(_get_settings_by_document(document.path)) not in run_types:
+        return
+
+    cell_diagnostics = _group_diagnostics_by_cell(await _lint_document_impl(document))
+
+    # Publish diagnostics for every code cell, replacing the previous diagnostics.
+    for cell_idx, cell in enumerate(notebook_document.cells):
+        if cell.kind is not NotebookCellKind.Code:
+            continue
```
dhruvmanila, line 483: We need to loop over each code cell here. To understand why, take for example a Notebook containing 2 cells:

1. On first save, both cell contains a single violation which got added using `notebookDocument/didOpen`
2. Suppose we fix the violation in the first cell and saved the notebook. This sends the `notebookDocument/didSave` request. Now, when we run Ruff on the updated document, we'll get diagnostics _only_ for the second cell. But, we need to clear diagnostics from the first cell which is why we need to loop over every cell.

ruff_lsp/server.py
```diff
@@ -366,7 +643,11 @@ def _get_severity(code: str) -> DiagnosticSeverity:
 
 @LSP_SERVER.feature(TEXT_DOCUMENT_HOVER)
 async def hover(params: HoverParams) -> Hover | None:
-    """LSP handler for textDocument/hover request."""
+    """LSP handler for textDocument/hover request.
+
+    This works for both Python files and Notebook Documents. For Notebook Documents,
+    the hover works at the cell level.
+    """
     document = LSP_SERVER.workspace.get_text_document(params.text_document.uri)
```
dhruvmanila, line 651: This is either a Python file or a Notebook cell.

ruff_lsp/server.py
```diff
@@ -528,6 +810,9 @@ async def code_action(params: CodeActionParams) -> list[CodeAction] | None:
     # Add "Ruff: Autofix" for every fixable diagnostic.
     if settings.get("codeAction", {}).get("fixViolation", {}).get("enable", True):
         if not params.context.only or CodeActionKind.QuickFix in params.context.only:
+            document = Document.from_text_document(
+                LSP_SERVER.workspace.get_text_document(params.text_document.uri)
+            )
```
dhruvmanila, line 815: Why are we creating the document again here when it's already constructed above at the start of the function?

This function is a generic implementation to create code actions at both source and line level. The document created above can only be used to create source level code actions as it represents either a Python file or an entire Notebook.

Here, we need a document which represents a Python file or an individual cell because we need to get the line level code action. Here, we already have the diagnostics information which is why this works as otherwise we would've been running Ruff on an individual cell.

ruff_lsp/server.py
```diff
@@ -736,47 +1023,116 @@ async def format_document(
...
 
 
-def _result_to_edits(
-    document: workspace.TextDocument,
-    result: RunResult | None,
-) -> list[TextEdit]:
+def _result_to_workspace_edit(
+    document: Document, result: RunResult | None
+) -> WorkspaceEdit | None:
+    """Converts a run result to a WorkspaceEdit."""
```
dhruvmanila, line 1058: Source level code actions should apply the fix for _every_ cell in a Notebook. Earlier the API was to create a single text edit to replace the entire content of a text document with the fixed source, create a workspace edit corresponding the text edit and send it back. This behavior is preserved when the document kind is Python.

But, for a Notebook, we need to create text edits for each cell where the fix exists which is why this was changed.

ruff_lsp/server.py
```diff
@@ -736,47 +1023,116 @@ async def format_document(
...
+    string with an assumption that the line endings are part of the strings itself.
+    """
+    if isinstance(fixed_source, list):
+        fixed_source = "".join(fixed_source)
+
+    new_source = _match_line_endings(original_source, fixed_source)
 
     # Skip last line ending in a notebook cell.
-    if document.uri.startswith("vscode-notebook-cell"):
+    if is_notebook_file:
```
dhruvmanila, line 1129: The server shouldn't rely on the URI format as then it wouldn't work for other editors.

ruff_lsp/server.py
```diff
@@ -736,47 +1023,116 @@ async def format_document(
...
         if new_source.endswith("\r\n"):
             new_source = new_source[:-2]
         elif new_source.endswith("\n"):
             new_source = new_source[:-1]
```
dhruvmanila, line 1133: This is more of a TODO for me to check if this is actually required now. Ruff adds and removes the newline at the end of every cell but I need to verify that behavior. I'm not sure what kind of example can be used here. I might have to check with a Ruff version when Jupyter Notebook support was not present.

ruff_lsp/settings.py
```diff
@@ -1,8 +1,19 @@
...
 from typing_extensions import Literal, TypedDict
 
-Run = Literal["onSave", "onType"]
+
+@enum.unique
+class Run(str, enum.Enum):
+    """When to run Ruff."""
+
+    OnType = "onType"
+    """Run Ruff on every keystroke."""
```
MichaReiser, line 13: Oh god, docstrings break my brain.... This is confusing after only having used languages where comments appear before the elements.

ruff_lsp/settings.py
```diff
@@ -94,8 +105,8 @@ def lint_args(settings: UserSettings) -> list[str]:
 def lint_run(settings: UserSettings) -> Run:
     """Get the `lint.run` setting from the user settings."""
     if "lint" in settings and "run" in settings["lint"]:
-        return settings["lint"]["run"]
+        return Run(settings["lint"]["run"])
```
MichaReiser, line 108: How is this change related to Jupyter notebook support? What motivates the change?

ruff_lsp/server.py
```diff
@@ -175,6 +204,149 @@ class VersionModified(NamedTuple):
...
+class Notebook(TypedDict):
+    """The JSON representation of a Notebook Document."""
+
+    metadata: Any
+    nbformat: int
+    nbformat_minor: int
+    cells: list[CodeCell | MarkdownCell]
+
+
+def _create_notebook_json(notebook_document: NotebookDocument) -> str:
```
MichaReiser, line 318: Oh wow, this feels kind of expensive. Do we run this on every keystroke or only on save? How does it perform for large notebooks?

ruff_lsp/server.py
MichaReiser: Might be worth adding this explanation to the inline comment

MichaReiser (approved):
LGTM. Would love to get a second look at this because I'm unfamiliar with our Python LSP implementation

ruff_lsp/server.py
```diff
@@ -3,20 +3,27 @@
...
 import asyncio
+import enum
 import json
 import logging
 import os
 import re
 import shutil
 import sys
 import sysconfig
+from collections.abc import Iterable, Mapping
```
charliermarsh, line 14: Just confirming that these are available on Python 3.7. Assuming yes given that CI passed.

ruff_lsp/server.py
```diff
@@ -175,6 +204,149 @@ class VersionModified(NamedTuple):
...
+        notebook_document = LSP_SERVER.workspace.get_notebook_document(cell_uri=uri)
+        if notebook_document is None:
+            notebook_document = LSP_SERVER.workspace.get_notebook_document(
+                notebook_uri=uri
+            )
+        if notebook_document is None:
+            text_document = LSP_SERVER.workspace.get_text_document(uri)
+            return cls.from_text_document(text_document)
+        else:
+            return cls.from_notebook_document(notebook_document)
```
charliermarsh, line 278: Maybe a bit clearer as:

```python
if notebook_document:
    return cls.from_notebook_document(notebook_document)

text_document = LSP_SERVER.workspace.get_text_document(uri)
return cls.from_text_document(text_document)
```

This keeps all the notebook logic up top, rather than separating it between lines 270 and then the jump to line 278.

ruff_lsp/server.py
```diff
@@ -175,6 +204,149 @@ class VersionModified(NamedTuple):
...
+
+    def is_stdlib_file(self) -> bool:
+        """Return True if the document belongs to standard library."""
+        return utils.is_stdlib_file(self.path)
+
+    def is_notebook_file(self) -> bool:
+        """Return True if the document belongs to a Notebook or a cell in a Notebook."""
+        return self.kind is DocumentKind.Notebook or self.path.endswith(".ipynb")
```
charliermarsh, line 286: Interesting, so the latter case is a cell? We use `DocumentKind.Text`?

ruff_lsp/server.py
charliermarsh: Do you mind adding something to this effect as a comment?

ruff_lsp/server.py
```diff
@@ -550,13 +835,16 @@ async def code_action(params: CodeActionParams) -> list[CodeAction] | None:
...
+                LSP_SERVER.workspace.get_text_document(params.text_document.uri)
+            )
             lines: list[str] | None = None
             for diagnostic in params.context.diagnostics:
                 if diagnostic.source == "Ruff":
                     noqa_row = cast(DiagnosticData, diagnostic.data).get("noqa_row")
                     if noqa_row is not None:
                         if lines is None:
-                            lines = document.lines
+                            lines = document.source.splitlines(keepends=True)
```
charliermarsh, line 847: Is this _definitely_ equivalent?

ruff_lsp/server.py
```diff
@@ -679,99 +968,166 @@ async def resolve_code_action(params: CodeAction) -> CodeAction:
...
-    text_document = LSP_SERVER.workspace.get_text_document(uri)
-    results = await _fix_document_impl(text_document)
-    LSP_SERVER.apply_edit(
-        _create_workspace_edits(text_document, results),
-        "Ruff: Fix all auto-fixable problems",
-    )
+    document = Document.from_uri(uri)
+    workspace_edit = await _fix_document_impl(document)
+    if workspace_edit is None:
+        return
```
charliermarsh, line 989: Is this a behavior change? Like does this action now _not_ show up in some cases?

ruff_lsp/server.py
```diff
@@ -679,99 +968,166 @@ async def resolve_code_action(params: CodeAction) -> CodeAction:
...
-        results = await _fix_document_impl(document)
-        params.edit = _create_workspace_edits(document, results)
+        params.edit = await _fix_document_impl(document)
 
     return params
 
 
 @LSP_SERVER.command("ruff.applyAutofix")
-async def apply_autofix(arguments: tuple[TextDocument]):
+async def apply_autofix(ls: server.LanguageServer, arguments: tuple[TextDocument]):
```
charliermarsh, line 984: How do we now have the `ls` here?

ruff_lsp/server.py
```diff
@@ -1192,13 +1539,9 @@ async def _run_check_on_document(
     )
 
 
-async def _run_format_on_document(document: workspace.TextDocument) -> RunResult | None:
+async def _run_format_on_document(document: Document) -> RunResult | None:
```
charliermarsh, line 1542: Wow, so this works interchangeably with Python files and notebooks now?

ruff_lsp/server.py
```diff
@@ -679,99 +968,166 @@ async def resolve_code_action(params: CodeAction) -> CodeAction:
...
-    LSP_SERVER.apply_edit(
-        _create_workspace_edits(text_document, results),
-        "Ruff: Format document",
-    )
+    document = Document.from_uri(uri)
+    results = await _run_format_on_document(document)
+    workspace_edit = _result_to_workspace_edit(document, results)
+    if workspace_edit is None:
+        return
+    LSP_SERVER.apply_edit(workspace_edit, "Ruff: Format document")
```
charliermarsh, line 1013: Can you walk me through what happens if we receive a notebook cell here? How does that interact with `_run_format_on_document`?

dhruvmanila:
Thanks for the reviews! I just realized while playing around with the `Format Cell` capability that the cell formatting doesn't work as it should've. The reason is that each cell is represented as a text document with the cell source and our abstraction will not construct a JSON representation for it.

There are 2 ways to go about solving this:
1. Implement the CLI flag as described here https://github.com/astral-sh/ruff/issues/6847 (`--extension ipynb=py`)
2. Add a new dedicated `DocumentKind.Cell` variant in the `Document` abstraction where the source will be the JSON representation containing a single cell. This will require some changes in translating the result into `TextEdit` to account for this new variant.

(Sorry for the churn, I must've missed this in my earlier testing and I wasn't aware of the `Format Cell` action 😓)

ruff_lsp/settings.py
dhruvmanila: Yeah, I agree but this is required for the editor to show the documentation :)

ruff_lsp/settings.py
dhruvmanila: With a new document support, the values are being used in multiple places. Earlier, a string literal was used directly but that means the value needs to be repeated in multiple places. The values being `"onType"`, `"onSave"`. The downsides of using a string literal is we've no way of knowing that a small type like `"ontype"` ("t" is lowercase) will break the code without running it. With an enum, the editor can provide auto-completion, diagnostics, and rename wherever they're being utilized.

ruff_lsp/server.py
dhruvmanila: Good questions. What about this function feels expensive? I guess the `json.dumps` part?

> Do we run this on every keystroke or only on save?

It depends on user settings but it could run on every keystroke when `"run": "onType"`.

> How does it perform for large notebooks?

I'll test it out.

We could potentially create our own layer of notebook synchronization but it feels not worth the effort if we're going to move the LSP to Rust.

Context: https://github.com/openlawlibrary/pygls/issues/394

ruff_lsp/server.py
dhruvmanila: Yes, I'm developing locally on 3.7 :)

ruff_lsp/server.py
dhruvmanila: Yes, although I'm thinking of adding it's own variant (https://github.com/astral-sh/ruff-lsp/pull/264#issuecomment-1782176658)

ruff_lsp/server.py
dhruvmanila: Are you asking w.r.t. the original implementation? If so, yes.

If you're asking w.r.t. the `document` variable, then yes as well because we're using the text document version for this case which is either a Python file or a Notebook cell.

ruff_lsp/server.py
dhruvmanila: > Is this a behavior change? 

No, it's not.

> Like does this action now not show up in some cases?

No, this action is registered and always shown on the client side using the VSCode extension. When invoked through VSCode, the code executes this command registered in the LSP.

ruff_lsp/server.py
dhruvmanila: Ah, it's not strictly required. We can either use the `ls` instance or the global `LSP_SERVER`, they both are the same.

ruff_lsp/server.py
dhruvmanila: Yes, the `Document` object creates an abstraction to store all the information required in the format (specifically the source) it is required.

ruff_lsp/server.py
dhruvmanila: The `applyFormat` command works at the Notebook level. 

For a Notebook the `arguments` is going to contain a single entry which will be a `TextDocument` for the _current_ cell (active cell or the cell containing the cursor). We'll construct a `Document` which represents either a Python file or a Notebook.

After the formatting is done, the `_result_to_workspace_edit` will generate a single `TextDocumentEdit` for the Python file or a list of `TextDocumentEdit` for each cell in  a Notebook. These are then a part of `WorkspaceEdit` which are then applied.

ruff_lsp/server.py
```diff
@@ -1137,22 +1563,28 @@ def _executable_version(executable: str) -> Version:
...
-async def _run_check_on_document(
-    document: workspace.TextDocument,
+class DocumentSource(NamedTuple):
+    """The source of a document."""
+
+    path: str
+    """The path to the document."""
+
+    text: str
+    """The text of the document."""
```
dhruvmanila, line 1573: This is so that we can pass a different source to Ruff than the one in the document. For example, in `Format Cell` document, we'll pass the notebook representation containing a single cell and then compare against the original source. Refer to `format_document` function.

ruff_lsp/server.py
dhruvmanila: I've reverted this to use the `pygls` implementation

dhruvmanila:
## Demo

### Formatting

* Using `Format Cell`
* Using `Ruff: Format Document`

https://github.com/astral-sh/ruff-lsp/assets/67177269/b83054f3-197f-4d8d-ae40-baefeaffdf9a

### Code Actions

* Fix individual code action
* Disable rule comment
* Fix all
* Format imports

https://github.com/astral-sh/ruff-lsp/assets/67177269/f29a7901-8799-4e67-b97f-c1dd3d6333aa

### Diagnostics

* `OnType`

https://github.com/astral-sh/ruff-lsp/assets/67177269/62e2b055-5368-4909-b29d-2b057f4811d1

* `OnSave`

https://github.com/astral-sh/ruff-lsp/assets/67177269/9819d5f4-1e8e-4800-bc66-a1055fcb16e1

### Hover

https://github.com/astral-sh/ruff-lsp/assets/67177269/f56103c2-c974-4b19-831d-9ee680cc9cf8

ruff_lsp/server.py
dhruvmanila: I've simplified this part such that each branch uses the document through our custom abstraction or the one from `pygls`.

charliermarsh:
(@dhruvmanila - FYI looks like a CI failure here.)

dhruvmanila:
> (@dhruvmanila - FYI looks like a CI failure here.)

Sorry, it's fixed.

DonnyVerduijn:
LGTM

ruff_lsp/server.py
```diff
@@ -175,6 +205,214 @@ class VersionModified(NamedTuple):
...
+    """The kind of document."""
+
+    Text = enum.auto()
+    """A Python file."""
+
+    Notebook = enum.auto()
+    """A Notebook Document."""
+
+    Cell = enum.auto()
+    """A cell in a Notebook Document."""
```
dhruvmanila, line 233: I had to introduce the cell specific document kind now that they're being used in multiple places.

ruff_lsp/server.py
```diff
@@ -175,6 +205,214 @@ class VersionModified(NamedTuple):
...
+        )
+
+    @classmethod
+    def from_cell_or_text_uri(cls, uri: str) -> Self:
+        """Create a `Document` representing either a Python file or a Notebook cell from
+        the given URI.
+
+        The function will try to get the Notebook cell first, and if there's no cell
+        with the given URI, it will fallback to the text document.
+        """
```
dhruvmanila, line 288: Not sure if there's any better name for this.

ruff_lsp/server.py
```diff
@@ -475,14 +830,17 @@ async def code_action(params: CodeActionParams) -> list[CodeAction] | None:
                 and len(params.context.only) == 1
                 and kind in params.context.only
             ):
-                edits = await _fix_document_impl(document, only="I001")
-                if edits:
+                workspace_edit = await _fix_document_impl(
+                    Document.from_cell_or_text_uri(params.text_document.uri),
+                    only="I001",
+                )
+                if workspace_edit:
```
dhruvmanila, line 837: Code actions now work at cell level for a Notebook. This is consistent with everything else in the VS Code ecosystem and a user can use builtin commands (`Ruff:`) to run an action on the entire notebook.

ruff_lsp/server.py
```diff
@@ -679,99 +1059,183 @@ async def resolve_code_action(params: CodeAction) -> CodeAction:
...
-        elif new_source.endswith("\n"):
-            new_source = new_source[:-1]
+def _result_single_cell_notebook_to_edits(
+    document: Document, result: RunResult
+) -> list[TextEdit] | None:
+    """Converts a run result to a list of TextEdits.
 
-    if new_source == document.source:
+    The result is expected to be a single cell Notebook Document.
+    """
```
dhruvmanila, line 1208: This function exists because for formatting provider, we need to return a list of `TextEdit` while for the command we need to apply a `WorkspaceEdit`.

dhruvmanila:
(@charliermarsh sorry for asking a review after approving, this is the final commit :))

I've pushed it as a separate commit to facilitate the review process: https://github.com/astral-sh/ruff-lsp/pull/264/commits/244a2461656724e5c4ef2b8e01c4ebf757148bdd

### Updates since the last review

The major update is that now the code actions work at cell level. Earlier, they were applied at the Notebook level. My thinking before was that source level code actions should be applied on the entire Notebook but it doesn't work well with VS Code settings, specifically the following:

```jsonc
{
  // Enable formatting the entire Notebook on save
  "notebook.formatOnSave.enabled": true,
  // Run the enabled code actions on the entire Notebook on save
  "notebook.codeActionsOnSave": {
    "source.fixAll": true,
    "source.organizeImports.ruff": true
  },
}
```

This also means that the generic `Organize Imports` / `Fix All` actions in VS Code works only for the current cell which is again consistent with the isort VS Code extension.

charliermarsh:
I'll review and approve this today, but before doing so I'll also check it out and poke around locally.

charliermarsh (approved):
Excellent, tested this out myself too.

PR #265, opened 2023-10-06
Use document path instead of an instance

dhruvmanila:
## Summary

The `TextDocument` instance will always exists, so we can remove the `| None`
from the type signature. The reason being that `pygls` will always return a
`TextDocument` instance via `workspace.get_text_document` method.

Also, the `document.path` field can never be `None` so the check is useless as
well.

### DIFF
--- a/ruff_lsp/server.py
+++ b/ruff_lsp/server.py
@@ -200,7 +200,7 @@ def did_close(params: DidCloseTextDocumentParams) -> None:
 async def did_save(params: DidSaveTextDocumentParams) -> None:
     """LSP handler for textDocument/didSave request."""
     document = LSP_SERVER.workspace.get_text_document(params.text_document.uri)
-    if lint_run(_get_settings_by_document(document)) in ("onType", "onSave"):
+    if lint_run(_get_settings_by_document(document.path)) in ("onType", "onSave"):
         diagnostics: list[Diagnostic] = await _lint_document_impl(document)
         LSP_SERVER.publish_diagnostics(document.uri, diagnostics)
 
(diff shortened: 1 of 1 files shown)


dhruvmanila:
* main
  * **PR #265** <a href="https://app.graphite.dev/github/pr/astral-sh/ruff-lsp/265" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/astral-sh/ruff-lsp/265?utm_source=stack-comment).

ruff_lsp/server.py
```diff
@@ -1024,16 +1023,11 @@ def _get_document_key(document: workspace.TextDocument) -> str | None:
     return None
 
 
-def _get_settings_by_document(
-    document: workspace.TextDocument | None,
-) -> WorkspaceSettings:
-    if document is None or document.path is None:
-        return list(WORKSPACE_SETTINGS.values())[0]
-
```
dhruvmanila, line 1032: This part becomes irrelevant as it's never executed.

ruff_lsp/server.py
```diff
@@ -1013,8 +1013,7 @@ def _update_workspace_settings(settings: list[WorkspaceSettings]) -> None:
             }
 
 
-def _get_document_key(document: workspace.TextDocument) -> str | None:
-    document_workspace = Path(document.path)
+def _get_document_key(document_workspace: Path) -> str | None:
```
charliermarsh, line 1016: Can we not pass a `Document` in here and then call `Path` on it internally? What's the motivation for passing in `Path`?

dhruvmanila: So, for notebook we have a different type `NotebookDocument`. The motivation being that the event handler will pass in either the notebook or file path along with the source code to the `_lint_document_impl`.

We could create a type `NotebookOrTextDocument` which would abstract out the `path` and `source` field since that's the only thing we use. Or, we could use a `Protocol`.

dhruvmanila: I'm actually thinking of using `str` here instead and use `Path` whenever required instead of instantiating it early on.

dhruvmanila: I'm actually thinking of using `str` here instead and use `Path` whenever required instead of instantiating it early on.

charliermarsh (approved):

ruff_lsp/server.py
charliermarsh: Got it, what you have here seems reasonable.

PR #266, opened 2023-10-07
Use `--output-format` instead of `--format`

zanieb:
`--format` is deprecated; includes backwards compatibility so we do not need to bump the linter version

### DIFF
--- a/ruff_lsp/server.py
+++ b/ruff_lsp/server.py
@@ -1172,7 +1172,7 @@ async def _run_check_on_document(
         else:
             argv.append(arg)
 
-    # If the Ruff version is sufficiently recent, use the deprecated `--format`
+    # If the Ruff version is not sufficiently recent, use the deprecated `--format`
     # argument instead of `--output-format`.
     if not VERSION_REQUIREMENT_OUTPUT_FORMAT.contains(
         executable.version, prereleases=True
--- a/ruff_lsp/server.py
+++ b/ruff_lsp/server.py
@@ -113,7 +113,10 @@ TOOL_DISPLAY = "Ruff"
 # Require at least Ruff v0.0.291 for formatting, but allow older versions for linting.
 VERSION_REQUIREMENT_FORMATTER = SpecifierSet(">=0.0.291,<0.2.0")
 VERSION_REQUIREMENT_LINTER = SpecifierSet(">=0.0.189,<0.2.0")
+# Version requirement for use of the "ALL" rule selector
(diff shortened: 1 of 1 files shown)


ruff_lsp/server.py
```diff
@@ -155,7 +155,7 @@ class VersionModified(NamedTuple):
     "--watch",
     # Arguments that are not supported at all, and will error when provided.
     # "--stdin-filename",
-    # "--format",
+    # "--output-format",
```
zanieb, line 158: https://github.com/astral-sh/ruff-lsp/pull/145/files#r1349400413

ruff_lsp/server.py
```diff
@@ -121,7 +121,7 @@ class VersionModified(NamedTuple):
     "--no-cache",
     "--no-fix",
     "--quiet",
-    "--format",
+    "--output-format",
```
charliermarsh, line 124: I think we need to do this in a backwards-compatible way, otherwise this will fail on all versions of Ruff prior to when that flag was introduced. We probably need to vary these arguments depending on the detected Ruff version?

zanieb: Yeah I went to bump the versions then realized backwards compatibility was probably better — I did this as you were reviewing :)

zanieb:
I want to add some tests to make sure this is backwards compatible — let's not merge yet

ruff_lsp/server.py
```diff
@@ -1167,6 +1170,15 @@ async def _run_check_on_document(
...
             argv.append(arg)
 
+    # If the Ruff version is sufficiently recent, use the deprecated `--format`
+    # argument instead of `--output-format`.
+    if not VERSION_REQUIREMENT_OUTPUT_FORMAT.contains(
+        executable.version, prereleases=True
+    ):
+        index = argv.index("--output-format")
+        argv.pop(index)
+        argv.insert(index, "--format=json")
```
charliermarsh, line 1180: Should the second arg here be `--format`, since `json` is provided as a separate entry?

zanieb: Yeah thanks — copilot inserted that -.-

charliermarsh (approved):

ruff_lsp/server.py
```diff
@@ -1169,6 +1172,15 @@ async def _run_check_on_document(
         else:
             argv.append(arg)
 
+    # If the Ruff version is sufficiently recent, use the deprecated `--format`
+    # argument instead of `--output-format`.
```
dhruvmanila, line 1176: ?
```suggestion
    # If the Ruff version is not sufficiently recent, use the deprecated `--format`
    # argument instead of `--output-format`.
```

PR #269, opened 2023-10-09
Add CI tests for the oldest supported Ruff version

zanieb:


zanieb:
Just a poc right now — I think I may go with a different approach

.github/workflows/ci.yaml
```diff
@@ -24,6 +25,9 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         run: just install
+      - name: Install lowest supported Ruff version
+        if: ${{ matrix.ruff == 'lowest' }}
+        run: pip install ruff==$(./get-oldest-supported-version.py ruff)
```
charliermarsh, line 30: I also wouldn't object to just hard-coding this and adding pointer comments between here and the `pyproject.toml` or similar.

zanieb:
#270 is nicer to me

PR #270, opened 2023-10-10
Add the latest and oldest Ruff versions to the CI test matrix

zanieb:


.github/workflows/ci.yaml
```diff
@@ -7,11 +7,38 @@ on:
     branches: [main]
 
 jobs:
+  ruff-versions:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install dependencies
+        run: |
+          sudo apt-get install -y curl jq ripgrep
```
konstin, line 16: curl and jq are preinstalled: https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md

zanieb: Oh cool!

.github/workflows/ci.yaml
```diff
@@ -7,12 +7,68 @@ on:
...
+          LATEST=$( \
+            curl -L \
+              -H "Accept: application/vnd.github+json" \
+              -H "X-GitHub-Api-Version: 2022-11-28" \
+              https://api.github.com/repos/astral-sh/ruff/releases/latest \
+              | jq '.tag_name' --raw-output \
+              | cut -c2- \
+          )
+          # Get the oldest supported version from the pyproject.toml
+          OLDEST=$(rg -No '"ruff>=(.*)"' -r '$1' pyproject.toml)
```
zanieb, line 47: It's way more painful to extract the first matching group like this without ripgrep, afaict

tests/test_server.py
```diff
@@ -16,11 +17,19 @@
 print(x)
 """
 
+VERSION_REQUIREMENT_ASTRAL_DOCS = Version("0.0.291")
+
 
-class TestServer(unittest.TestCase):
+class TestServer:
```
zanieb, line 23: I need the pytest fixture to be accessible here

zanieb: I'd like to follow-up with a pull request that removes the class, but want to keep the diff focused here.

tests/test_server.py
```diff
@@ -149,7 +164,7 @@ def _handler(params):
                         {
                             "code": "F401",
                             "codeDescription": {
-                                "href": "https://docs.astral.sh/ruff/rules/unused-import"
+                                "href": expected_docs_url + "rules/unused-import"
```
charliermarsh, line 167: I wish we could just store "fixtures by version" or something, rather than having to construct these and encode facts like "Which URL is present at which version?" in the test logic. Like, in Rust, we'd just run the command with different versions and snapshot the output, which feels easier to maintain and understand. I don't know if I have a concrete suggestion here -- more reacting to the increased logic in the tests now.

.github/workflows/ci.yaml
charliermarsh: I would honestly be fine hard-coding this _and_ the latest version, since we already search-and-replace the latest version every time we upgrade the LSP. Fewer dependencies, easier to maintain (IMO). But of course I defer to you.

charliermarsh (approved):

konstin (approved):

.github/workflows/ci.yaml
zanieb: Maybe it's just me but a sed invocation like that is entirely incomprehensible haha :)

I'd like to try this out since I wrote it up and haven't tried generating matrix test versions like this before, but at the first sign of trouble I'm happy to just switch to hard-coded versions.

tests/test_server.py
zanieb: Hm interesting yeah.. this whole bit would make more sense as a snapshot too.

A strong case for us building Python test tooling <3

PR #275, opened 2023-10-16
Provide rule doc links

### LINKED ISSUE #274: Provide a link to rule doc using `textDocument/documentLink`
Is this more discoverable than hover?

https://github.com/astral-sh/ruff-lsp/assets/17039389/193d7fae-c0a2-43ec-93df-a8e580bc4ddb

<details><summary>Code change</summary>
<p>

```diff
diff --git a/ruff_lsp/server.py b/ruff_lsp/server.py
index ea7df2a..a1eb35a 100755
--- a/ruff_lsp/server.py
+++ b/ruff_lsp/server.py
@@ -23,6 +23,7 @@ from lsprotocol.types import (
     TEXT_DOCUMENT_DID_OPEN,
     TEXT_DOCUMENT_DID_SAVE,
     TEXT_DOCUMENT_FORMATTING,
+    TEXT_DOCUMENT_DOCUMENT_LINK,
     TEXT_DOCUMENT_HOVER,
     AnnotatedTextEdit,
     ClientCapabilities,
@@ -39,6 +40,8 @@ from lsprotocol.types import (
     DidOpenTextDocumentParams,
     DidSaveTextDocumentParams,
     DocumentFormattingParams,
+    DocumentLink,
+    DocumentLinkParams,
     Hover,
     HoverParams,
     InitializeParams,
@@ -397,6 +400,59 @@ async def hover(params: HoverParams) -> Hover | None:
     return None
 
 
+@LSP_SERVER.feature(TEXT_DOCUMENT_DOCUMENT_LINK)
+async def document_link(params: DocumentLinkParams) -> list[DocumentLink] | None:
+    """LSP handler for textDocument/documentLink request."""
+    document = LSP_SERVER.workspace.get_document(params.text_document.uri)
+
+    links: list[DocumentLink] = []
+    for line_index, line in enumerate(document.lines):
+        match = NOQA_REGEX.search(line)
+        if not match:
+            continue
+
+        codes = match.group("codes")
+        if not codes:
+            continue
+
+        codes_start = match.start("codes")
+        for match in CODE_REGEX.finditer(codes):
+            start, end = match.span()
+            start += codes_start
+            end += codes_start
+            code = match.group()
+            result = await _run_subcommand_on_document(
+                document,
+                VERSION_REQUIREMENT_LINTER,
+                args=[
+                    "--explain",
+                    code,
+                    "--format",
+                    "json",
+                ],
+            )
+            log_to_output(f"result.stdout: {result.stdout}")
+            if result.stdout:
+                name = json.loads(result.stdout.decode("utf-8")).get("name")
+                links.append(
+                    DocumentLink(
+                        range=Range(
+                            start=Position(
+                                line=line_index,
+                                character=start,
+                            ),
+                            end=Position(
+                                line=line_index,
+                                character=end,
+                            ),
+                        ),
+                        target=f"https://docs.astral.sh/ruff/rules/{name}",
+                    )
+                )
+
+    return links
+
+
 ###
 # Code Actions.
 ###
```

</p>
</details>

harupy:
## Summary

Close #274

## Test Plan

Manually tested

### DIFF
--- a/ruff_lsp/server.py
+++ b/ruff_lsp/server.py
@@ -23,6 +23,7 @@ from lsprotocol.types import (
     TEXT_DOCUMENT_DID_OPEN,
     TEXT_DOCUMENT_DID_SAVE,
     TEXT_DOCUMENT_FORMATTING,
+    TEXT_DOCUMENT_DOCUMENT_LINK,
     TEXT_DOCUMENT_HOVER,
     AnnotatedTextEdit,
     ClientCapabilities,
@@ -39,6 +40,8 @@ from lsprotocol.types import (
     DidOpenTextDocumentParams,
     DidSaveTextDocumentParams,
     DocumentFormattingParams,
+    DocumentLink,
+    DocumentLinkParams,
     Hover,
     HoverParams,
     InitializeParams,
@@ -364,25 +367,33 @@ NOQA_REGEX = re.compile(
 CODE_REGEX = re.compile(r"[A-Z]+[0-9]+")
 
 
-@LSP_SERVER.feature(TEXT_DOCUMENT_HOVER)
-async def hover(params: HoverParams) -> Hover | None:
-    """LSP handler for textDocument/hover request."""
-    document = LSP_SERVER.workspace.get_text_document(params.text_document.uri)
-    match = NOQA_REGEX.search(document.lines[params.position.line])
+def _parse_codes(line: str) -> list[tuple(str, int, int)]:
+    match = NOQA_REGEX.search(line)
     if not match:
-        return None
+        return []
 
     codes = match.group("codes")
     if not codes:
-        return None
+        return []
 
+    res = []
     codes_start = match.start("codes")
     for match in CODE_REGEX.finditer(codes):
         start, end = match.span()
         start += codes_start
         end += codes_start
+        code = match.group()
+        res.append((code, start, end))
+    return res
+
+
+@LSP_SERVER.feature(TEXT_DOCUMENT_HOVER)
+async def hover(params: HoverParams) -> Hover | None:
+    """LSP handler for textDocument/hover request."""
+    document = LSP_SERVER.workspace.get_text_document(params.text_document.uri)
+
+    for code, start, end in _parse_codes(document.lines[params.position.line]):
         if start <= params.position.character < end:
-            code = match.group()
             result = await _run_subcommand_on_document(
                 document, VERSION_REQUIREMENT_LINTER, args=["--explain", code]
             )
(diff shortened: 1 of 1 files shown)


ruff_lsp/server.py
```diff
@@ -397,6 +408,35 @@ async def hover(params: HoverParams) -> Hover | None:
...
+                links.append(
+                    DocumentLink(
+                        range=Range(
+                            start=Position(line=index, character=start),
+                            end=Position(line=index, character=end),
+                        ),
+                        tooltip=f"Jump to {code} documentation",
+                        target=f"https://docs.astral.sh/ruff/rules/{name}",
+                    )
+                )
```
charliermarsh, line 435: Does this iterate over the entire document, checking every line?

harupy: Yes, it does...

ruff_lsp/server.py
```diff
@@ -364,25 +367,33 @@ def _get_severity(code: str) -> DiagnosticSeverity:
 CODE_REGEX = re.compile(r"[A-Z]+[0-9]+")
 
 
-@LSP_SERVER.feature(TEXT_DOCUMENT_HOVER)
-async def hover(params: HoverParams) -> Hover | None:
-    """LSP handler for textDocument/hover request."""
-    document = LSP_SERVER.workspace.get_text_document(params.text_document.uri)
-    match = NOQA_REGEX.search(document.lines[params.position.line])
+def _parse_codes(line: str) -> list[tuple(str, int, int)]:
```
dhruvmanila, line 370: Could we have a `NamedTuple` / `dataclass` / `TypedDict` instead of `(str, int, int)`? It's hard to know what those represents otherwise.

ruff_lsp/server.py
```diff
@@ -397,6 +408,35 @@ async def hover(params: HoverParams) -> Hover | None:
     return None
 
 
+@LSP_SERVER.feature(TEXT_DOCUMENT_DOCUMENT_LINK)
+async def document_link(params: DocumentLinkParams) -> list[DocumentLink] | None:
+    """LSP handler for textDocument/documentLink request."""
+    document = LSP_SERVER.workspace.get_document(params.text_document.uri)
+
+    links: list[DocumentLink] = []
+    for index, line in enumerate(document.lines):
```
dhruvmanila, line 417: Oh, this request applies for an _entire_ document and not for a specific range. It makes me wonder how often this request is being sent by the client. If it's being sent multiple times then we should think of ways to optimize this as otherwise this might take some time for a large file irrespective of whether it contains the codes or not.

A few options might be:
* Applying the regex on the entire document using [`re.finditer`](https://docs.python.org/3/library/re.html?highlight=re%20finditer#re.finditer) instead of applying the regex on each line
* Using [`documentLink/resolve`](https://docs.python.org/3/library/re.html?highlight=re%20finditer#re.finditer) to lazily compute the rule name. So, for `textDocument/documentLink` request, we would only compute the range and when the client explicitly requests for the link, we would run the command and update the `documentLink` object.

harupy: @dhruvmanila Thanks for the comment!

> It makes me wonder how often this request is being sent by the client.

- When a file is opened.
- When a file is saved.

harupy: `textDocument/documentLink` is sent on type.

harupy: I'll close this :) Iterating over the entire document on type is too expensive.

PR #283, opened 2023-10-17
Require `ruff>=0.1.0`

dhruvmanila:
## Summary

This PR updates the minimum required version for Ruff to be `v0.1.0`. This is
because that version includes the `cell` field which will be used for Jupyter
Notebook support.


dhruvmanila:
* main
  * **PR #283** <a href="https://app.graphite.dev/github/pr/astral-sh/ruff-lsp/283" target="_blank">Graphite</a>  👈
    * **PR #284** <a href="https://app.graphite.dev/github/pr/astral-sh/ruff-lsp/284" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/astral-sh/ruff-lsp/283?utm_source=stack-comment).

zanieb:
Can we avoid this? It'd be preferable to support an older version still and just toggle this behavior depending on the Ruff version (as we do for several other features)

dhruvmanila:
I guess it could be avoided but then the Notebook support will be available only if the user has v0.1.0.

zanieb:
That seems preferable to me, but might be worth some more opinions. We opted for that approach in #266

dhruvmanila:
Implementation is simple because it's just a `dict.get("cell")`.

PR #286, opened 2023-10-17
Run tests against the main branch of Ruff

zanieb:
Adds tests against ruff@main, installed from source.

The test matrix is kind of big here, we could drop Python versions or operating systems (for a subset of the matrix) but I'm not sure what subset is best. Since this repository is relatively calm, it might not hurt to just have a bunch of checks.

Since this depends on us opening a pull request here to notice breakage with ruff, I'm also adding https://github.com/astral-sh/ruff/pull/8016 which runs tests on each change to ruff.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -86,8 +86,11 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
           cache: 'pip'
+
       - name: Install dependencies
-        run: just install
+        run: |
+          pip install -U pip
+          just install
 
       - name: Install test Ruff version from PyPI
         if: ${{ matrix.ruff-version != env.RUFF_UNRELEASED_REF }} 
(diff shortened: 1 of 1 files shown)


konstin:
imho we should split the test matrix into running the test across all platforms and python versions and testing special cases which only run on select targets, such as min python and latest python on ubuntu-latest.

zanieb:
@konstin I'm happy to make that change. Should we test latest or main across all targets?

konstin:
I'd say latest, i'd special case main, though it's not a strong preference. If we make main first then we should run CI on every commit of astra-sh/ruff first, though my main reason for latest over main is practicality, `pip install ruff` is much faster than a whole release build of ruff.

zanieb:
Great I agree!

zanieb:
I've made the requested changes to the matrix — interestingly `ruff version` is not working in the pip install...

zanieb:
but it works locally... bigger hmm

zanieb:
Forcing reinstall was the key — presumably because we don't do fancy versioning for development commits

konstin (approved):

PR #306, opened 2023-11-01
feat(ci): add `cache-dependency-path`

SauravMaheshkar:
## Summary

This PR adds the `cache-dependency-path` parameter which allows us to explicitly specify files which contain packages. More information can be found on the [docs for actions/setup-python](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#caching-packages)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -22,6 +22,10 @@ jobs:
         with:
           python-version: "3.11"
           cache: 'pip'
+          cache-dependency-path: |
+            pyproject.toml
+            requirements.txt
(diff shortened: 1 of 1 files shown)


SauravMaheshkar:
> Thanks for contributing! Would you mind pushing an empty commit so we can verify this actually changes the number of cached packages during installation?

Makes sense, just pushed

SauravMaheshkar:
In the "Install dependencies" step, we can see that it uses the cached version of every package (`Using cached ...`)

zanieb:
Thank you!

zanieb (approved):

PR #310, opened 2023-11-07
Fix exclusive rules selection for `only` arg

dhruvmanila:
## Summary

Earlier, if a user had define a rule selection or exclusion using the `lint.args` setting, it would collide with the `only` parameter. The purpose of the `only` parameter is to only run Ruff for the given rule excluding everything else. The "excluding everything else" part was done with `--extend-ignore=ALL`. The problem, as highlighted in the linked issue, was that if there's a user defined `--ignore` / `--select` (or their `--extend-*` version), then the `ALL` directive won't work because:

> When breaking ties between enabled and disabled rules (via select and ignore, respectively), more specific prefixes override less specific prefixes.
>
> _Reference: https://docs.astral.sh/ruff/settings/#ignore_

This means that between `--select=E,F` and `--ignore=ALL`, the former wins.

This PR fixes this issue by checking for the following argument pattern in user defined settings, skipping them if the `only` parameter is given:

* `["--select", "A,B,C"]` / `["--select=A,B,C"]`
* `["--ignore", "A,B,C"]` / `["--ignore=A,B,C"]`

And their `--extend-*` version.

## Test Plan

Manual testing locally :)

fixes: https://github.com/astral-sh/ruff-vscode/pull/328

### DIFF
--- a/ruff_lsp/server.py
+++ b/ruff_lsp/server.py
@@ -1157,11 +1157,30 @@ async def _run_check_on_document(
     executable = _find_ruff_binary(settings, VERSION_REQUIREMENT_LINTER)
     argv: list[str] = CHECK_ARGS + list(extra_args)
 
+    skip_next_arg = False
     for arg in lint_args(settings):
+        if skip_next_arg:
+            skip_next_arg = False
+            continue
         if arg in UNSUPPORTED_CHECK_ARGS:
             log_to_output(f"Ignoring unsupported argument: {arg}")
-        else:
-            argv.append(arg)
+            continue
+        # If we're trying to run a single rule, we need to make sure to skip any of the
+        # arguments that would override it.
+        if only:
+            # Case 1: Option and it's argument as separate items
+            # (e.g. `["--select", "F821"]`).
+            if arg in ("--select", "--extend-select", "--ignore", "--extend-ignore"):
+                # Skip the following argument assuming it's a list of rules.
+                skip_next_arg = True
+                continue
+            # Case 2: Option and it's argument as a single item
+            # (e.g. `["--select=F821"]`).
+            elif arg.startswith(
+                ("--select=", "--extend-select=", "--ignore=", "--extend-ignore=")
+            ):
+                continue
+        argv.append(arg)
 
     # If the Ruff version is not sufficiently recent, use the deprecated `--format`
     # argument instead of `--output-format`.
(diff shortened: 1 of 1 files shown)


dhruvmanila:
* main
  * **PR #310** <a href="https://app.graphite.dev/github/pr/astral-sh/ruff-lsp/310" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/astral-sh/ruff-lsp/310?utm_source=stack-comment).

ruff_lsp/server.py
```diff
@@ -1157,10 +1157,27 @@ async def _run_check_on_document(
...
 
+    skip_user_arg = False
     for arg in lint_args(settings):
+        if skip_user_arg:
+            skip_user_arg = False
+            continue
         if arg in UNSUPPORTED_CHECK_ARGS:
             log_to_output(f"Ignoring unsupported argument: {arg}")
         else:
+            if only:
```
zanieb, line 1168: The check for unsupported seems like it could use `continue`?

ruff_lsp/server.py
```diff
@@ -1157,10 +1157,27 @@ async def _run_check_on_document(
...
+                if arg in (
+                    "--select",
+                    "--extend-select",
+                    "--ignore",
+                    "--extend-ignore",
+                ):
+                    skip_user_arg = True
+                    continue
+                elif arg.startswith(
```
zanieb, line 1177: Why do you use `startswith` here? This check looks the same as the one above?

ruff_lsp/server.py
dhruvmanila: Are you referring to the other `if` branch? I might be misunderstanding but I don't think it does as it's a `if ... else ...`?

ruff_lsp/server.py
dhruvmanila: For the following conditions:
1. `["--select", "A,B,C"]`
2. `["--select=A,B,C"]`

Thanks for checking, I'll add this as inline documentation

ruff_lsp/server.py
zanieb: Like

```
        if arg in UNSUPPORTED_CHECK_ARGS:
            log_to_output(f"Ignoring unsupported argument: {arg}")
            continue
       if only:
            ...
```

It doesn't look like there's any reason to use an `else` and indent all this code.

dhruvmanila: Oh interesting! Yeah, this looks far more readable. I'll update the code, thanks!

ruff_lsp/server.py
```diff
@@ -1157,11 +1157,35 @@ async def _run_check_on_document(
...
+                "--ignore",
+                "--extend-ignore",
+            ):
+                # Skip the following argument assuming it's a list of rules.
+                skip_next_arg = True
+                continue
+            # Case 2: Option and it's argument as a single item
+            # (e.g. `["--select=F821"]`).
+            elif arg.startswith(
+                ("--select", "--extend-select", "--ignore", "--extend-ignore")
```
charliermarsh, line 1185: Should we add `=` to the end of these, to be stricter?

PR #314, opened 2023-11-09
Avoid removing all contents when formatting file with syntax error

zanieb:
Exploring a resolution to https://github.com/astral-sh/ruff-vscode/issues/336

It looks like at https://github.com/astral-sh/ruff-lsp/compare/v0.0.42...v0.0.43#diff-9cc81288343e2240bb3387d88ab41ebf706e40c86af24cf96bad8919a48d5cceL762-L763 we no longer exit early if stdout is empty. This pull request does not restore that logic, but rather adds a return code check and exits if there is a bad return code.

## Summary

- Add failing test case for formatting code with a syntax error
- Track return code of `ruff format`
- Do not format documents with non-zero return code from Ruff


zanieb:
Remaining work involves:

- Determine if the lint fixes are affected and need a similar fix
- Add logging of stdout from Ruff on bad return codes
- Test in VS Code

zanieb:
We could consider applying an additional safe guard, but I'm not sure when return codes are intentionally non-zero

```diff
diff --git a/ruff_lsp/server.py b/ruff_lsp/server.py
index e1e6091..437827f 100755
--- a/ruff_lsp/server.py
+++ b/ruff_lsp/server.py
@@ -1142,7 +1142,7 @@ def _result_to_workspace_edit(
     document: Document, result: RunResult | None
 ) -> WorkspaceEdit | None:
     """Converts a run result to a WorkspaceEdit."""
-    if result is None:
+    if result is None or result.exit_code != 0:
         return None
 
     if document.kind is DocumentKind.Text:
```

ruff_lsp/server.py
```diff
@@ -1108,8 +1112,9 @@ async def format_document(params: DocumentFormattingParams) -> list[TextEdit] |
     # request itself can only act on a text document. A cell in a Notebook is
     # represented as a text document.
     document = Document.from_cell_or_text_uri(params.text_document.uri)
+
     result = await _run_format_on_document(document)
-    if result is None:
+    if result is None or result.exit_code != 0:
```
charliermarsh, line 1117: I think we should also add this to `_lint_document_impl` for good measure.

zanieb: I think the patch in https://github.com/astral-sh/ruff-lsp/pull/314#issuecomment-1804875791 would apply to all uses

charliermarsh: Ahh you're right, we might exit 1 on lint errors.

charliermarsh: (In fact pretty sure we do.)

charliermarsh (approved):

Issue #267, opened 2023-10-07
Formatting and "Fix All" doesn't work if it removes entire file content


dhruvmanila:
If you take the following code example in VSCode:

```python
import os
```

And run the command `Ruff: Fix all auto-fixable problems`, nothing will happen.

But, suppose you add an additional statement:

```python
import os

print("hello world")
```

Now, if you run the same command, it'll remove the import.

The reason it's not working in the first case is because we create a `WorkspaceEdit` only if the fixed source code isn't empty. But, in the first case it's going to be empty.

A new statement isn't even needed, just add some extra newlines and you'll be able to reproduce the bug.

dhruvmanila:
(This change was reverted in #317)

dhruvmanila:
Using https://github.com/astral-sh/ruff/pull/8596, we can now safely write empty fixes on supported Ruff version.

PR #319, opened 2023-11-10
Always write empty fixes on Ruff >= 0.1.6

charliermarsh:
## Summary

We can avoid the safeguard we introduced in https://github.com/astral-sh/ruff-lsp/pull/317 for newer Ruff versions, since Ruff will now avoid writing empty output for excluded files (see: https://github.com/astral-sh/ruff/pull/8596).

Closes https://github.com/astral-sh/ruff-lsp/issues/267.

## Test Plan

I tested this with a local debug build. Now, when I "Fix all" on a non-excluded file with `import os`, it _does_ properly get removed, while running "Format" or "Fix all" on an excluded file leaves the contents untouched.

### DIFF
--- a/ruff_lsp/server.py
+++ b/ruff_lsp/server.py
@@ -140,10 +140,10 @@ TOOL_DISPLAY = "Ruff"
 # Require at least Ruff v0.0.291 for formatting, but allow older versions for linting.
 VERSION_REQUIREMENT_FORMATTER = SpecifierSet(">=0.0.291,<0.2.0")
 VERSION_REQUIREMENT_LINTER = SpecifierSet(">=0.0.189,<0.2.0")
-# Version requirement for use of the "ALL" rule selector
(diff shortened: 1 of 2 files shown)


ruff_lsp/server.py
```diff
@@ -140,10 +140,10 @@ class VersionModified(NamedTuple):
 # Require at least Ruff v0.0.291 for formatting, but allow older versions for linting.
 VERSION_REQUIREMENT_FORMATTER = SpecifierSet(">=0.0.291,<0.2.0")
 VERSION_REQUIREMENT_LINTER = SpecifierSet(">=0.0.189,<0.2.0")
-# Version requirement for use of the "ALL" rule selector
-VERSION_REQUIREMENT_ALL_SELECTOR = SpecifierSet(">=0.0.198,<0.2.0")
```
charliermarsh, line 144: Turns out this is now unused.

dhruvmanila (approved):

PR #341, opened 2023-12-14
Avoid erasing files when Ruff fails with an error

charliermarsh:
## Summary

A user reported an issue in the VS Code extension (https://github.com/astral-sh/ruff/issues/9123) in which their files were being erased on-save. The issue is that we don't check the exit code of all commands, and so if `ruff check` fails with an _error_, then we end up wiping the file entirely.

This PR adds exit code-checking for all commands in the LSP. Further, we now surface such failures to the user directly (see bottom right):

Screen Shot 2023-12-13 at 10 57 20 PM

Closes https://github.com/astral-sh/ruff/issues/9123

### DIFF
--- a/ruff_lsp/server.py
+++ b/ruff_lsp/server.py
@@ -558,6 +558,14 @@ async def _lint_document_impl(document: Document) -> list[Diagnostic]:
     result = await _run_check_on_document(document)
     if result is None:
         return []
+
+    # For `ruff check`, 0 indicates successful completion with no remaining
+    # diagnostics, 1 indicates successful completion with remaining diagnostics,
+    # and 2 indicates an error.
+    if result.exit_code not in (0, 1):
+        show_error(f"Ruff: Lint failed ({result.stderr.decode('utf-8')})")
+        return []
+
     return _parse_output(result.stdout) if result.stdout else []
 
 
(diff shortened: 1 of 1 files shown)


ruff_lsp/server.py
```diff
@@ -558,6 +558,14 @@ async def _lint_document_impl(document: Document) -> list[Diagnostic]:
     result = await _run_check_on_document(document)
     if result is None:
         return []
+
+    # For `ruff check`, 0 indicates successful completion with no remaining
+    # diagnostics, 1 indicates successful completion with remaining diagnostics,
+    # and 2 indicates an error.
+    if result.exit_code != 0 and result.exit_code != 1:
```
MichaReiser, line 565: Nit: I don't know what's more idiomatic Python
```suggestion
    if not result.exit_code in [0, 1]:
```

ruff_lsp/server.py
```diff
@@ -1141,6 +1159,17 @@ async def _fix_document_impl(
         extra_args=["--fix"],
         only=only,
     )
+
+    if result is None:
```
MichaReiser, line 1163: When does `ruff` return `None`? if there were no fixes? Is this an error state (same for the other commands)

MichaReiser (approved):

ruff_lsp/server.py
dhruvmanila: 😬 
```python
if result.exit_code not in [0, 1]:
	...
```

ruff_lsp/server.py
charliermarsh: `_run_check_on_document` will return `None` if the _extension_ chooses to ignore the file (for example, if it appears to be a stdlib file in the current environment).

ruff_lsp/server.py
charliermarsh: I think `not in (0, 1)` is most idiomatic, probably.

MichaReiser: > 😬
> 
> ```python
> if result.exit_code not in [0, 1]:
> 	...
> ```

Python's hard

charliermarsh: Haha both work!

dhruvmanila (approved):
Thanks for the quick fix! It works on Notebook as expected.

PR #343, opened 2023-12-17
Only surface errors when stderr is present

charliermarsh:
This has the side effect of avoiding syntax errors, for which we shouldn't show a pop-up.


charliermarsh:
This is not a good solution, but it's the best solution I have right now. This will now surface command-line errors (like non-existent arguments), but not syntax errors.

PR #350, opened 2024-01-04
Fix stale link to Ruff fix docs

vEnhance:
## Summary

One-line repair of a stale URL in the README

## Test Plan

N/A because this is a documentation change only.


zanieb (approved):
Thanks!

PR #351, opened 2024-01-05
Support notebook-scoped VS Code actions

### LINKED ISSUE #320: Ruff/Ruff VScode deletes imports if they aren't used in the same cell
To reproduce:

Python version: 3.11.1
Ruff version: 0.1.5
Ruff-vscode version: v2023.46.0

VScode settings.json:
```
{
  "editor.rulers": [119],
  "editor.defaultFormatter": "esbenp.prettier-vscode",
  "editor.formatOnSave": true,
  "[python]": {
    "editor.formatOnSave": true,
    "editor.codeActionsOnSave": {
      "source.fixAll": true,
      "source.organizeImports": true
    },
    "editor.defaultFormatter": "charliermarsh.ruff"
  },
  "notebook.codeActionsOnSave": {
    "source.fixAll": true,
    "source.organizeImports": true
  },
  "explorer.confirmDragAndDrop": false,
  "github.copilot.enable": {
    "*": true,
    "plaintext": false,
    "markdown": true,
    "scminput": false
  },
  "jupyter.askForKernelRestart": false,
  "jupyter.widgetScriptSources": ["jsdelivr.com", "unpkg.com"],
  "[sql]": {
    "editor.defaultFormatter": "inferrinizzard.prettier-sql-vscode"
  },
  "editor.minimap.enabled": false,
  "Prettier-SQL.keywordCase": "upper"
}
```

Any notebook will do. E.g. in cell 1:

```python
from pathlib import Path
```

In cell 2:

```python
Path.cwd()
```

Saving this notebook will cause `from pathlib import Path` to be deleted from the first cell. Commenting out `"source.fixAll": true,` in the notebook.codeActionsOnSave bit of settings.json stops this but obviously also turns off Ruff's functionality.

It's possible I've done something wrong or this is expected behaviour but just flagging it in case it's useful. Really amazing how much you've been shipping lately!

charliermarsh:
## Summary

VS Code added support for `notebook.*`-scoped code actions (https://code.visualstudio.com/api/references/vscode-api#CodeActionKind), which are intended to run over the entire document, unlike the actions that omit the `notebook.` namespace, which instead only operate over individual cells. The latter are problematic for actions that need global context, e.g., our unused imports rules.

When the `notebook.*` actions are triggered, VS Code sends down the first cell (see: https://github.com/microsoft/vscode/issues/193120) as the URI. This PR adds handlers for the actions and logic to use the notebook, rather than the cell.

Closes #320.

## Test Plan

Used the following `settings.json`:

```json
{
  "[python]": {
      "editor.defaultFormatter": "charliermarsh.ruff",
      "editor.codeActionsOnSave": {
        "source.organizeImports.ruff": true
      }
    },
    "notebook.codeActionsOnSave": {
      "notebook.source.fixAll": true
    }
}
```

Verified that imports used across cells were not removed (whereas `"source.fixAll": true` _did_ cause imports to be removed).

### DIFF
--- a/ruff_lsp/server.py
+++ b/ruff_lsp/server.py
@@ -204,6 +204,22 @@ UNSUPPORTED_FORMAT_ARGS = [
     # "--stdin-filename",
 ]
 
+# Standard code action kinds, scoped to Ruff.
+SOURCE_FIX_ALL_RUFF = f"{CodeActionKind.SourceFixAll.value}.ruff"
+SOURCE_ORGANIZE_IMPORTS_RUFF = f"{CodeActionKind.SourceOrganizeImports.value}.ruff"
+
+# Notebook code action kinds.
+NOTEBOOK_SOURCE_FIX_ALL = f"notebook.{CodeActionKind.SourceFixAll.value}"
+NOTEBOOK_SOURCE_ORGANIZE_IMPORTS = (
+    f"notebook.{CodeActionKind.SourceOrganizeImports.value}"
+)
+
+# Notebook code action kinds, scoped to Ruff.
+NOTEBOOK_SOURCE_FIX_ALL_RUFF = f"notebook.{CodeActionKind.SourceFixAll.value}.ruff"
+NOTEBOOK_SOURCE_ORGANIZE_IMPORTS_RUFF = (
+    f"notebook.{CodeActionKind.SourceOrganizeImports.value}.ruff"
+)
+
 
 ###
 # Document
@@ -804,11 +820,19 @@ class LegacyFix(TypedDict):
     TEXT_DOCUMENT_CODE_ACTION,
     CodeActionOptions(
         code_action_kinds=[
+            # Standard code action kinds.
             CodeActionKind.QuickFix,
             CodeActionKind.SourceFixAll,
             CodeActionKind.SourceOrganizeImports,
-            f"{CodeActionKind.SourceFixAll.value}.ruff",
-            f"{CodeActionKind.SourceOrganizeImports.value}.ruff",
+            # Standard code action kinds, scoped to Ruff.
+            SOURCE_FIX_ALL_RUFF,
+            SOURCE_ORGANIZE_IMPORTS_RUFF,
+            # Notebook code action kinds.
+            NOTEBOOK_SOURCE_FIX_ALL,
+            NOTEBOOK_SOURCE_ORGANIZE_IMPORTS,
+            # Notebook code action kinds, scoped to Ruff.
+            NOTEBOOK_SOURCE_FIX_ALL_RUFF,
+            NOTEBOOK_SOURCE_ORGANIZE_IMPORTS_RUFF,
         ],
         resolve_provider=True,
     ),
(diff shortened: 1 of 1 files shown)


ruff_lsp/server.py
```diff
@@ -1063,17 +1086,14 @@ async def resolve_code_action(params: CodeAction) -> CodeAction:
...
+    if (
+        settings["organizeImports"]
+        and params.kind == CodeActionKind.SourceOrganizeImports
     ):
         # Generate the "Ruff: Organize Imports" edit
         params.edit = await _fix_document_impl(document, only=["I001", "I002"])
 
-    elif settings["fixAll"] and params.kind in (
-        CodeActionKind.SourceFixAll,
-        f"{CodeActionKind.SourceFixAll.value}.ruff",
```
charliermarsh, line 1075: I'm pretty sure we only ever pass the full action here? This is for deferred actions.

ruff_lsp/server.py
```diff
@@ -821,26 +829,44 @@ async def code_action(params: CodeActionParams) -> list[CodeAction] | None:
...
+
+    def document(uri: str, kind: str) -> Document:
+        if kind in (
+            # For `notebook`-scoped actions, use the Notebook Document instead of
+            # the cell, despite being passed the URI of the first cell.
+            # See: https://github.com/microsoft/vscode/issues/193120
+            f"notebook.{CodeActionKind.SourceFixAll.value}",
+            f"notebook.{CodeActionKind.SourceOrganizeImports.value}",
+            f"notebook.{CodeActionKind.SourceFixAll.value}.ruff",
+            f"notebook.{CodeActionKind.SourceOrganizeImports.value}.ruff",
```
zanieb, line 841: Should these be constants defined elsewhere so they can't ever mismatch the ones in `CodeActionOptions`?

ruff_lsp/server.py
```diff
@@ -821,26 +829,44 @@ async def code_action(params: CodeActionParams) -> list[CodeAction] | None:
     first, and if there's no cell with the given URI, it will fallback to the text
     document.
     """
+
+    def document(uri: str, kind: str) -> Document:
```
zanieb, line 833: Can we call this something else? e.g.`document_from` or `load_document` or `document_from_kind` or `get_document` 

I mistook it for a variable instead of a function later on.

ruff_lsp/server.py
```diff
@@ -1063,17 +1091,14 @@ async def resolve_code_action(params: CodeAction) -> CodeAction:
 
     settings = _get_settings_by_document(document.path)
 
-    if settings["organizeImports"] and params.kind in (
-        CodeActionKind.SourceOrganizeImports,
-        f"{CodeActionKind.SourceOrganizeImports.value}.ruff",
```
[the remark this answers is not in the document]
zanieb: Is this what you're commenting on in https://github.com/astral-sh/ruff-lsp/pull/351/files#r1442483542?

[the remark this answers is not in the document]
charliermarsh: Yeah, I don't see how this would ever end up triggered here.

zanieb (approved):
Sweet fix!

PR #361, opened 2024-01-24
Add support for disabling the linter

charliermarsh:
## Summary

This PR enables users to specify `"ruff.lint.enable": false` to use the LSP (or extension) solely for formatting.

"Organize Imports" is still supported even when the linter is disabled.

See: https://github.com/astral-sh/ruff-vscode/issues/329.

## Test Plan

- Ran the VS Code extension.
- Verified that diagnostics appeared for unused imports.
- Set `"ruff.lint.enable": false`.
- Verified that the diagnostics disappeared.
- Ran "Organize Imports".
- Verified that the imports were sorted.
- Ran "Fix All".
- Verified that the imports (which were unused) were not removed.

### DIFF
--- a/justfile
+++ b/justfile
@@ -9,8 +9,8 @@ install:
   pip install --no-deps -r requirements-dev.txt
 
 fmt:
-  ruff check --fix ./ruff_lsp ./tests
   ruff format ./ruff_lsp ./tests
+  ruff check --fix ./ruff_lsp ./tests
 
 check:
   ruff check ./ruff_lsp ./tests
--- a/ruff_lsp/server.py
+++ b/ruff_lsp/server.py
@@ -79,6 +79,7 @@ from ruff_lsp.settings import (
     UserSettings,
     WorkspaceSettings,
     lint_args,
+    lint_enable,
     lint_run,
 )
 from ruff_lsp.utils import RunResult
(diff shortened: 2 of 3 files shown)


ruff_lsp/server.py
```diff
@@ -935,8 +962,10 @@ def document_from_kind(uri: str, kind: str) -> Document:
 
     actions: list[CodeAction] = []
 
-    # Add "Ruff: Autofix" for every fixable diagnostic.
-    if settings.get("codeAction", {}).get("fixViolation", {}).get("enable", True):
+    # If the linter is enabled, add "Ruff: Autofix" for every fixable diagnostic.
+    if settings.get("codeAction", {}).get("fixViolation", {}).get(
+        "enable", True
+    ) and lint_enable(settings):
```
dhruvmanila, line 968: nit: slight preference to have `lint_enable` first which then reads as "if lint is enabled, and if this code action is enabled, ..."

dhruvmanila (approved):
Looks good, thanks!

ruff_lsp/server.py
charliermarsh: Agreed, thanks.

justfile
```diff
@@ -9,8 +9,8 @@ install:
   pip install --no-deps -r requirements-dev.txt
 
 fmt:
-  ruff check --fix ./ruff_lsp ./tests
   ruff format ./ruff_lsp ./tests
+  ruff check --fix ./ruff_lsp ./tests
```
MichaReiser, line 13: What's the reason for changing the order here? Fixing should run before formatting to avoid that fixes introduce new formatting errors. Or did you run into situations where the formatter produced new lint violations, if so, that's a bug (except if it is a rule that this known to be incompatible with the formatter)

charliermarsh: I was getting annoyed because the formatter wouldn't run if any checks failed (like E501). But the right fix is to use `--fix-only`, changing...

(This setup is a vestige from before the formatter existed.)

charliermarsh: https://github.com/astral-sh/ruff-lsp/pull/369

Issue #116, opened 2023-05-19
Investigate range formatting


konstin:
Figure out and document which capabilities to use for range formatting and what kind of offsets we get from VS Code and other LSP clients. Make a plan how to pass those to ruff

PR #373, opened 2024-01-31
Range formatting support

MichaReiser:
## Summary

This PR implements support for range formatting for Python files. It requires https://github.com/astral-sh/ruff/pull/9733

Note to myself: Writing Python is hard :sweat_smile: 

Closes https://github.com/astral-sh/ruff-lsp/issues/116
Closes https://github.com/astral-sh/ruff-vscode/issues/319

## Limitations

Range formatting Notebooks isn't supported because it raises a few interesting questions:

* Should the range be relative to the concatenated cell content? 
* Should the range use a custom notation `cell:offset`
* What if the range falls on the boundary of two cells? 

I feel like implementing range formatting will be straightforward with a Rust LSP because:
* We can make the assumption that only ever a single cell gets formatted (solves some of the problems above)
* We can directly call into the range formatting API without having to find the proper CLI abstractions.

I tried to teach VS Code that Range formatting isn't supported for notebooks by using a `textSelector` but VS code doesn't seem to care :(
That's why the implementation defaults to format the entire cell for now. I'm open to suggestions.

## Test Plan

* Added tests
[Screencast from 2024-01-31 16-11-40.webm](https://github.com/astral-sh/ruff-lsp/assets/1203881/0d935068-b93e-4714-af10-4128aafe58d5)

### DIFF
--- a/ruff_lsp/server.py
+++ b/ruff_lsp/server.py
@@ -1215,14 +1215,41 @@ async def apply_format(arguments: tuple[TextDocument]):
 
 
 @LSP_SERVER.feature(TEXT_DOCUMENT_FORMATTING)
-async def format_document(
-    params: DocumentFormattingParams, range: Range | None = None
+async def format_document(params: DocumentFormattingParams) -> list[TextEdit] | None:
+    return await _format_document_impl(params, None)
+
+
+@LSP_SERVER.feature(
+    TEXT_DOCUMENT_RANGE_FORMATTING,
+    DocumentRangeFormattingRegistrationOptions(
+        document_selector=[
+            TextDocumentFilter_Type1(language="python", scheme="file"),
+            TextDocumentFilter_Type1(language="python", scheme="untitled"),
+        ],
+        ranges_support=False,
+        work_done_progress=False,
+    ),
+)
+async def format_document_range(
+    params: DocumentRangeFormattingParams,
+) -> list[TextEdit] | None:
+    return await _format_document_impl(
+        DocumentFormattingParams(
+            params.text_document, params.options, params.work_done_token
+        ),
+        params.range,
+    )
+
+
+async def _format_document_impl(
+    params: DocumentFormattingParams, range: Range | None
 ) -> list[TextEdit] | None:
     # For a Jupyter Notebook, this request can only format a single cell as the
     # request itself can only act on a text document. A cell in a Notebook is
     # represented as a text document. The "Notebook: Format notebook" action calls
     # this request for every cell.
     document = Document.from_cell_or_text_uri(params.text_document.uri)
+
     settings = _get_settings_by_document(document.path)
 
     # We don't support range formatting of notebooks yet but VS Code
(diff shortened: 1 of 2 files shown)


tests/test_format.py
```diff
@@ -70,6 +72,56 @@ async def test_format_code_with_syntax_error(tmp_path, ruff_version: Version):
...
+    document = Document.from_text_document(workspace.get_text_document(uri))
+    settings = _get_settings_by_document(document.path)
+
+    handle_unsupported = (
+        pytest.raises(RuntimeError, match=f"Ruff .* required, but found {ruff_version}")
+        if not VERSION_REQUIREMENT_RANGE_FORMATTING.contains(ruff_version)
+        else nullcontext()
+    )
+
+    with handle_unsupported:
```
AlexWaygood, line 113: It just comes down to stylistic preference, really, but you could consider using an [`ExitStack`](https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack) here instead of the `nullcontext()` ternary:

```suggestion
    with ExitStack() as stack;
        if not VERSION_REQUIREMENT_RANGE_FORMATTING.contains(ruff_version):
            stack.enter_context(
                pytest.raises(
                    RuntimeError, match=f"Ruff .* required, but found {ruff_version}"
                )
            )

```

ruff_lsp/server.py
```diff
@@ -1840,6 +1880,25 @@ async def _run_format_on_document(
         document.path,
     ]
 
+    if format_range:
+        # Convert the start and end to character offsets instead of line:column
+        lines = document.source.splitlines(True)
+        codec = PositionCodec(PositionEncodingKind.Utf16)
+        range = codec.range_from_client_units(lines, format_range)
+        start = format_range.start.character
```
charliermarsh, line 1888: Should / could this be `range.start.character?`

charliermarsh (approved):

ruff_lsp/server.py
```diff
@@ -1209,15 +1216,21 @@ async def apply_format(arguments: tuple[TextDocument]):
...
     # request itself can only act on a text document. A cell in a Notebook is
     # represented as a text document. The "Notebook: Format notebook" action calls
     # this request for every cell.
     document = Document.from_cell_or_text_uri(params.text_document.uri)
     settings = _get_settings_by_document(document.path)
 
-    result = await _run_format_on_document(document, settings)
+    # We don't support range formatting of notebooks yet but VS Code
+    # doesn't seem to respect the document filter. For now, format the entire cell.
+    range = None if document.kind is DocumentKind.Cell else range
```
dhruvmanila, line 1231: This is unfortunate because as a user I wouldn't expect that "Format selection" in Jupyter notebook would format an entire cell. I'm not sure what the right solution is but I'd prefer to send a info / warning message to the user stating that range formatting isn't yet supported for Jupyter Notebooks instead.

dhruvmanila (approved):

tests/test_format.py
MichaReiser: Uhm, I just copied what the other tests did. I'll leave it for now for consistency

ruff_lsp/server.py
MichaReiser: I agree but I wasn't able to come up with something better. I'm worried that a warning would be rather annoying and force users to change their setting from `modificationsIfAvailable` to the entire document or they get annoyed by the warning on every save

tests/test_format.py
AlexWaygood: Sure, makes sense!

PR #374, opened 2024-02-01
Narrow Text edits 

MichaReiser:
## Summary

Try to narrow text edits to changed lines only. The implemented algorithm is very basic. It only trims the common suffix and postfix (line based) but doesn't do any diffing beyond that. 
The idea is to have a fast (with predictable performance) algorithm that at least avoids transmitting the entire formatted file when doing range formatting. 

## Test Plan

I played with range formatting, code fixes, and normal formatting in the VS code extension to verify that it doesn't truncate text.

### DIFF
--- a/ruff_lsp/server.py
+++ b/ruff_lsp/server.py
@@ -16,7 +16,6 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import Any, List, NamedTuple, Sequence, Union, cast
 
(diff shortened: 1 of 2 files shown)


charliermarsh (approved):

dhruvmanila (approved):
LGTM! I think you'd need to rebase to get the version requirement update.

PR #379, opened 2024-02-05
Update requirements.txt

MichaReiser:
## Summary

Run `just lock`

## Motivation

Running `pip install` locally fails with 

```
ERROR: Cannot install -r requirements.txt (line 35) and lsprotocol==2023.0.1 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested lsprotocol==2023.0.1
    pygls 1.2.1 depends on lsprotocol==2023.0.0

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
WARNING: There was an error checking the latest version of pip.
(.venv) 
```

(It doesn't matter if I use python 3.11 or 3.7)

The lspprotocol dependency was updated by Dependabot in https://github.com/astral-sh/ruff-lsp/pull/368

I don't know what I'm doing here is right (Python packaging is confusing...), but I feel like `pip install` should work 😆 

## Test Plan

Running `pip install` no longer fails

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -41,9 +41,9 @@ iniconfig==2.0.0 \
     --hash=sha256:2d91e135bf72d31a410b17c16da610a82cb55f6b0477d1a902134b24a455b8b3 \
     --hash=sha256:b6a85871a79d2e3b22d2d1b94ac2824226a63c6b741c88f7ae975f18b6778374
     # via pytest
-lsprotocol==2023.0.1 \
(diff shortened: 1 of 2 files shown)


charliermarsh (approved):

MichaReiser:
Okay I think I know what it is. CI and `just install` use `pip install --no-deps`... That solves the issue

PR #386, opened 2024-02-17
style: add None for empty return statements in server.py

mrKazzila:
## Summary

Simply added `None` to empty return to maintain consistency with other methods/functions that return None.
Example:

1. https://github.com/astral-sh/ruff-lsp/blob/bae2b2488b09ae3dc1b411ce845db780738827ee/ruff_lsp/server.py#L622
2. https://github.com/astral-sh/ruff-lsp/blob/bae2b2488b09ae3dc1b411ce845db780738827ee/ruff_lsp/server.py#L733
3. …

Also aligned with [PEP 8 recommendations](https://peps.python.org/pep-0008/#programming-recommendations) "Be consistent in return statements …".

## Test Plan

N/A because both `return` and `return None` return None in the same way.

### DIFF
--- a/ruff_lsp/server.py
+++ b/ruff_lsp/server.py
@@ -449,7 +449,7 @@ async def did_open(params: DidOpenTextDocumentParams) -> None:
     )
     settings = _get_settings_by_document(document.path)
     if not lint_enable(settings):
-        return
+        return None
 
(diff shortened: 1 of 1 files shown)


charliermarsh (approved):

charliermarsh:
Thanks!

PR #392, opened 2024-02-29
Use `ruff check` instead of `ruff`

MichaReiser:
## Summary

Use `ruff check` instead of `ruff` for linting. 

## Test Plan

I verified that: 

* ruff-lsp uses `ruff ...` for linting before making the change
* ruff-lsp uses `ruff check ...` after my change
* that diagnostics are still shown after my change


ruff_lsp/server.py
```diff
@@ -154,6 +154,7 @@ class VersionModified(NamedTuple):
 
 # Arguments provided to every Ruff invocation.
 CHECK_ARGS = [
+    "check",
```
charliermarsh, line 157: I assume this existed from `v0.0.189` onward?

PR #449, opened 2024-06-19
Avoid disable rule comment code action for `E999`

dhruvmanila:
## Summary

This PR updates the `ruff-lsp` server to avoid generating the "disable for this line" code action for `E999` (syntax error). This is a follow-up to https://github.com/astral-sh/ruff/pull/11901 which does the same for the new server.

## Test Plan

Run it locally and make sure there are no quick fixes in the VS Code menu:

Screenshot 2024-06-19 at 07 31 55
Screenshot 2024-06-19 at 07 31 40

### DIFF
--- a/ruff_lsp/server.py
+++ b/ruff_lsp/server.py
@@ -1016,7 +1016,7 @@ async def code_action(params: CodeActionParams) -> list[CodeAction] | None:
             )
             lines: list[str] | None = None
             for diagnostic in params.context.diagnostics:
-                if diagnostic.source == "Ruff":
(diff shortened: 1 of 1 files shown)


MichaReiser (approved):

dhruvmanila:
Note: Do not merge this PR right now, it needs to be clubbed with the 0.5 release mainly to match the behavior with `ruff server`.

dhruvmanila:
This is not required now since the `noqa_row` field in the JSON output will be `null` in case of syntax errors with https://github.com/astral-sh/ruff/pull/11901

PR #522, opened 2025-02-09
Avoid deprecation warning if client is VS Code

dhruvmanila:
## Summary

Related: https://github.com/astral-sh/ruff-vscode/pull/685

This PR removes the deprecation warning for VS Code client and removes the version check as well as it's not needed for other editors.

## Test Plan

While testing https://github.com/astral-sh/ruff-vscode/pull/685 with local `ruff-lsp` on this branch, VS Code should only provides notification from the extension and not the one from `ruff-lsp`.

### DIFF
--- a/ruff_lsp/server.py
+++ b/ruff_lsp/server.py
@@ -1616,10 +1616,7 @@ def initialize(params: InitializeParams) -> None:
 
     _update_workspace_settings(settings)
 
-    if (
-        params.client_info is not None
-        and params.client_info.name == "Visual Studio Code"
-    ):
+    if os.getenv("LS_SHOW_DEPRECATION_WARNING") == "False":
         # The extension is responsible for providing the deprecation warning.
         return
 
--- a/ruff_lsp/server.py
+++ b/ruff_lsp/server.py
@@ -1616,23 +1616,23 @@ def initialize(params: InitializeParams) -> None:
 
     _update_workspace_settings(settings)
 
-    # Use the Ruff executable from the first workspace to determine when to show the
-    # deprecation warning.
-    executable = _find_ruff_binary(
-        next(iter(WORKSPACE_SETTINGS.values())), version_requirement=None
-    )
+    if (
+        params.client_info is not None
+        and params.client_info.name == "Visual Studio Code"
+    ):
+        # The extension is responsible for providing the deprecation warning.
+        return
 
-    if VERSION_REQUIREMENT_NATIVE_SERVER.contains(executable.version, prereleases=True):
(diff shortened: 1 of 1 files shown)


ruff_lsp/server.py
```diff
@@ -1616,24 +1616,24 @@ def initialize(params: InitializeParams) -> None:
...
     _update_workspace_settings(settings)
 
-    # Use the Ruff executable from the first workspace to determine when to show the
-    # deprecation warning.
-    executable = _find_ruff_binary(
-        next(iter(WORKSPACE_SETTINGS.values())), version_requirement=None
+    if (
+        params.client_info is not None
+        and params.client_info.name == "Visual Studio Code"
+    ):
```
MichaReiser, line 1622: Does that work for other VS code like editors, e.g. https://vscodium.com/ which is popular on linux distros that only bundle open source code.

dhruvmanila: Oh, good point. I just checked, it uses a different client name (obviously). I'll update it and check for other editors that use the VS Code Marketplace.

dhruvmanila: I think it might just better to set an internal-only environment variable via the VS Code extension which is read in `ruff-lsp` to signal that the warnings are handled by the extension because ChatGPT gives me a long list:

```
Here’s a list of editors that support installing extensions from the VS Code Marketplace or OpenVSX Registry:

Editors Using the VS Code Marketplace
These editors have direct or unofficial access to Microsoft's VS Code Marketplace:

Visual Studio Code (Official) – Direct access
Visual Studio Code Insiders – Direct access
VSCodium (requires a workaround; Microsoft restricts Marketplace access)
Onivim 2 (planned support but not officially released)
JetBrains IDEs (via the [VSCode Extension Marketplace plugin](https://plugins.jetbrains.com/plugin/17029-vscode-extension-marketplace), limited support)
Editors Using OpenVSX Registry
These editors support OpenVSX, an open-source alternative to the VS Code Marketplace:

VSCodium – Default extension source
Eclipse Theia – Fully supports OpenVSX
Code-OSS (the open-source base of VS Code) – Configurable to use OpenVSX
Gitpod – Uses OpenVSX
Coder (formerly "code-server") – Uses OpenVSX
OpenVSX-compatible VS Code forks – Any editor that builds on Code-OSS can use OpenVSX
```

dhruvmanila: Added `LS_SHOW_DEPRECATION_WARNING`

MichaReiser (approved):

PR #529, opened 2025-08-08
chore(ci): address CI linting findings

woodruffw:
## Summary

This uses pinact and zizmor to burn down some CI issues. It also enables Dependabot for GitHub Actions for future automatic actions bumps.

(Pretty low priority I figure, since this repo is deprecated. But I figure it doesn't hurt to have the CI be in good order.)

## Test Plan

See what happens in CI.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -5,3 +5,14 @@ updates:
     schedule:
       interval: "daily"
       time: "12:00"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      time: "12:00"
+    cooldown:
+      default-days: 7
+    groups:
+      github-actions:
+        patterns:
+          - "*"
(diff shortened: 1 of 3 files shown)


.github/dependabot.yaml
```diff
@@ -5,3 +5,14 @@ updates:
     schedule:
       interval: "daily"
       time: "12:00"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
MichaReiser, line 11: Should we change this to monthly? I know, it's a trade-off with getting security fixes quickly, but we don't plan to make any new changes in this repository or releases, and clicking through renovate PRs also takes some time (every week).

MichaReiser (approved):
Thank you. 

I'm inclined to just disable CI entirely, given that we don't plan on making any changes to ruff-lsp or publish new releases. I guess it has the downside that it becomes more work if we have to publish a release (although it's not clear to me why that would be). 

Either way, let's focus on minimizing overhead. I'm fine merging this but I'd probably decrease the renovate frequency to reduce the noise this project generates.

PR #541, opened 2025-10-15
Migrate from deprecated `macos-13` runner to `macos-14-intel`

zanieb:


MichaReiser:
Thank you. But I'm not sure it's worth investing time into `ruff-lsp`. I currently don't see any reason why we should cut a release.

